PR #2, opened 2015-05-08
Add a few minor tweaks

tianon:
- verify downloaded file PGP signatures (since signatures are available)
- reorder instructions a little for better cache overlap (shared layers between the two versions)
- utilize `WORKDIR` to make `run.sh` slightly simpler and allow the JSON syntax for `CMD` which avoids an extra `sh` process
- utilize `exec` in `run.sh` to avoid a second extra `sh` process (which makes `java` become "pid 1" in the container, instead of the shell -- see `docker top <running-container>`)

### DIFF
--- a/4.5.4/Dockerfile
+++ b/4.5.4/Dockerfile
@@ -1,7 +1,6 @@
 FROM java:openjdk-8u45-jre
 MAINTAINER David Gageot <david.gageot@sonarsource.com>
 
-ENV SONAR_VERSION 4.5.4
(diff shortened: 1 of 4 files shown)


dgageot:
Thanks a lot

tianon:
:+1: thanks for submitting an image!  :heart:

dgageot:
I'm going to tweak the Dockerfile a little bit more. I'd like to get rid of the run.sh and make the command an entry point.

dgageot:
Might not be able to do both because of env. variables

PR #4, opened 2015-05-30
Let the user set JVM and launch parameters

glanzi:
I am running SonarQube behind a reverse proxy and thus that i need a simple solution to set the system property to change the web context:
docker run --name sonarqube -p 9000:9000 --env JAVA_OPTS=-Dsonar.web.context=/sonarqube sonarqube:5.1

### DIFF
--- a/4.5.4/run.sh
+++ b/4.5.4/run.sh
@@ -5,10 +5,10 @@ if [ "${1:0:1}" != '-' ]; then
 	exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+exec java $JAVA_OPTS -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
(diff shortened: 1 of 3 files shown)


dgageot:
Did you try passing the additional parameters to the docker command line? It's supposed to work. Please tell me if it doesn't.

```
docker run --name sonarqube -p 9000:9000 sonarqube:5.1 -Dsonar.web.context=/sonarqube
```

glanzi:
Thanks for your answer. It works when i set the java environment property in the docker command parameter or in my case in docker-compose.yml like "command: -Dsonar.web.context=/sonar".

dgageot:
good to hear! :-)

Batmat:
Hi,
I know it's not really a question related to SonarQube, but I'm interesting in using that trick to pass parameters to Sonar, but using Docker Compose.

Compose lets you pass env variables in a std way, but can't find anything else to do what's above other than overriding the "command" with the additional params. 
We don't really like it because we have to copy/paste the existing CMD in the Dockerfile. So that's more a workaround than a solution in our opinion.

Do you know of a better way to do that with Compose ?

Thanks a lot.

dgageot:
@Batmat sorry but no.

PR #9, opened 2015-08-11
Add SonarQube 5.1.2

Godin:


Godin:
@dgageot we forgot PR for docker-library/official-images ?

Godin:
For the record I created docker-library/official-images#967

PR #10, opened 2015-10-27
Fix readme links

mpailloncy:
Small fixes in README.

Otherwise, dunno why there isn't any image in version 4.5.5 in DockerHub ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ docker-sonarqube
 # Supported tags and respective Dockerfile links
 
 * `5.1`, `latest` ([Dockerfile](https://github.com/SonarSource/docker-sonarqube/blob/master/5.1/Dockerfile))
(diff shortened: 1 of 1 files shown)


Godin:
Thank you for contribution @mpailloncy !

The reason why there is no 4.5.5 image in Official Docker Library ( https://hub.docker.com/_/sonarqube/ ) is that this repository is just a "source" (melting pot) and separate PR to https://github.com/docker-library/official-images/blob/master/library/sonarqube is required for actual publishing.

Same is applicable for README here - it is unused and real content is generated from https://github.com/docker-library/docs/tree/master/sonarqube

I will change README here to explain this and make things more clearer. As well as creation of PRs to include SonarQube 4.5.6 and 5.2 to Docker Library.

Thus, I'm closing this PR without merge, but anyway thanks again for raising a question.

Godin:
README cleaned up in 2f7cc2f6ef7f0206b697c37df09fe2e0fd36c8f4

PR #12, opened 2015-11-28
Use links instead of hacking

tanarurkerem:
It is not necessary to expose postgresql port on host machine. Postgres will be accessible for the linked container(s) only.
The docker-compose starts postgres before sonar start.


Godin:
@tanarurkerem Thank you for contribution!

PR #13, opened 2016-01-04
Recipe for plugins

mOuLiNeX:
With this recipe, you can restart sonarqube (or postgresql) without loosing data and upgrade or install some plugins.


Godin:
@mOuLiNeX Thank you for contribution!

PR #18, opened 2016-02-11
dont stick with old version of jdk

lkwg82:
Changed version of of jdk to  latest 8. Due to ssl problems I had to replace curl with wget.

If this is fine to you, plz change the other versions as well!

I rate this is as structural security improvement.

Cheers Lars

### DIFF
--- a/5.3/Dockerfile
+++ b/5.3/Dockerfile
@@ -1,4 +1,4 @@
-FROM java:openjdk-8u45-jdk
+FROM java:8
 
 MAINTAINER David Gageot <david.gageot@sonarsource.com>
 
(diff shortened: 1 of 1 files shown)


Godin:
@lkwg82 could you please elaborate more about replacement of curl by wget?

Godin:
java image switched from Debian sid to jessie - https://github.com/docker-library/openjdk/commit/2695acad7bb0cd43353dc6c7d5cfc6424ab12b54
and seems that certificates are broken in it - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708
curl can perform downloading after downgrade of ca-certifcates package as described in
https://github.com/4commerce-technologies-AG/meteor/issues/37#issuecomment-184310053
and since wget is able to download using broken ca-certificates, this raises a question - is it safe to use it?

I also have doubts whether it is good idea to blindly use tag "8", which points on different minor versions in different moments of time, instead of explicit updates up to concrete version. For example we can safely upgrade to image "java:openjdk-8u66-jdk", which does not contain problem with ca-certificates.

lkwg82:
**regarding: ssl certificate**

maybe just wait before implement some downstream workarounds. I dont want to add a workaround with the comment ```// remove when bug XY is resolved```. IMHO this is a blocker for this PR.

**regarding: java version**

Is your contract really java:8-u45 or is it just java version 8. I really doubt you rely on specific patch level. Else you would miss any security updates or need to test each version again.
So in my opinion ```java:8``` is very different from versionless like ```java:latest```. This seems to me the best balance between *security* and *stability*.

Godin:
It has been quite some time without updates in upstream, while would be nice to resolve this, even if the only option (as per https://github.com/docker-library/openjdk/pull/40#issuecomment-216576620) is a temporary workaround - 77b12b561f3ed8ef6308a9d2e1adb25e17704b69

Anyway thank you for your contribution, we appreciate it :thumbsup:

PR #20, opened 2016-03-03
Remove sonar.web.javaAdditionalOpts from run.sh to allow proxy definition

resilva87:
Hi guys,

After trying desperately for a week to correctly set the proxy definition in sonar.properties (http://stackoverflow.com/questions/35709604/sonarqube-5-3-download-plugins-behind-proxy-https/35726595?noredirect=1#comment59208893_35726595) I've found out that defining `sonar.web.javaAdditionalOpts` in `run.sh` was the culprit to problem.

It seems that when the application is starting, it'll read `sonar.properties` file but it'll not merge or even replace `sonar.web.javaAdditionalOpts` from `run.sh`, in which case I could not set the required proxies for my environment.

The property in my `sonar.properties` was set like this: `sonar.web.javaAdditionalOpts=-Djava.security.egd=file:/dev/./urandom -Dhttps.proxyHost=proxy.XXX -Dhttps.proxyPort=YYY -Dhttp.proxyHost=proxy.XXX -Dhttp.proxyPort=YYY`

I believe this is a bug as it does not allow anyone to properly use proxies in their environment.

Cheers.

### DIFF
--- a/4.5.6/Dockerfile
+++ b/4.5.6/Dockerfile
@@ -34,7 +34,7 @@ RUN set -x \
 	&& rm sonarqube.zip* \
 	&& rm -rf $SONARQUBE_HOME/bin/*
 
-VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions"]
+VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions", "$SONARQUBE_HOME/logs"]
 
 WORKDIR $SONARQUBE_HOME
 COPY run.sh $SONARQUBE_HOME/bin/
--- a/4.5.6/run.sh
+++ b/4.5.6/run.sh
@@ -10,5 +10,4 @@ exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
   -Dsonar.jdbc.password="$SONARQUBE_JDBC_PASSWORD" \
   -Dsonar.jdbc.url="$SONARQUBE_JDBC_URL" \
-  -Dsonar.web.javaAdditionalOpts="-Djava.security.egd=file:/dev/./urandom" \
 	"$@"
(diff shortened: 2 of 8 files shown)


Godin:
@resilva87 I don't think that complete removal of default `-Djava.security.egd=file:/dev/./urandom` is a good idea. Maybe you have a better idea? Also I'm actually surprised that specification of `http.proxyHost` in `conf/sonar.properties` doesn't work, this sounds like a bug in upstream - in SonarQube.

resilva87:
@Godin Actually in this context you'are correct: `http.proxy{Host/Port}` does work. But not the `https` variant. Some plugins in Update Center require connection to `https://sonarsource.bintray.com` which fails even if `https.proxy{Host/Port}` are defined in `sonar.properties`. The only way I found, based [on](http://stackoverflow.com/questions/31296136/update-center-behind-proxy-howto-with-https-bintray-repository) [these](http://stackoverflow.com/questions/20495468/sonarqube-proxy-configuration-tricky), was setting both  `http` and `https` proxies in `sonar.web.javaAdditionalOpts`, which is working.

Godin:
Seems that indeed `https.proxyHost` and `https.proxyPort` can't be loaded from `sonar.properties` - see https://github.com/SonarSource/sonarqube/blob/5.3/sonar-core/src/main/java/org/sonar/core/util/DefaultHttpDownloader.java#L168-L170
And I created ticket - https://jira.sonarsource.com/browse/SONAR-7429 , feel free to vote.

Godin:
@resilva87 according to @simonbrandhof ( https://jira.sonarsource.com/browse/SONAR-7429 ) :
> "http.proxyUser" and "http.proxyPassword" are indeed properties defined by SonarQube. They're badly named as they're used for both HTTP and HTTPS. That deserves documentation to be updated.

did you tried this?

Godin:
@resilva87 oups, sorry, my bad - this statement doesn't include `proxyHost` and `proxyPort`

Godin:
@resilva87 also please check PR #22 - it allows to pass `sonar.web.javaAdditionalOpts` without removal of default value `-Djava.security.egd=file:/dev/./urandom`, e.g.
```
docker run --rm --name sonarqube -p 9000:9000 -p 9092:9092 \
  -e SONARQUBE_WEB_JVM_OPTS="-Xmx32m" \
  sonarqube
```

Godin:
Closed in preference to #22

PR #21, opened 2016-03-08
Add SonarQube 5.4

Godin:


Godin:
This also requires merge of docker-library/official-images#1525

Issue #17, opened 2016-01-18
Ability to set the sonar web context by environment 


rschlick:
Hi,

I need to change the default web context of sonar to /sonar. Is it possible to set this value through an docker environment variable ? 

Thanks,

Romain Schlick

rschlick:
I have proposed a pull request, but I don't have access on your repository.
I have clone your repository and push the commit on my repository here:
https://github.com/rschlick/docker-sonar

Godin:
`sonar.web.context` has been removed in SonarQube version 5.4 - see https://jira.sonarsource.com/browse/SONAR-7122

tobilarscheid:
I vote for reopening this, as the web context will be added in sonar 5.6 again!

https://jira.sonarsource.com/browse/SONAR-7122

Godin:
@tobilarscheid could you please check #22 - in comparison with this request it introduces more generic variable which can be used to specify various options via `sonar.web.javaAdditionalOpts` and not just context, and in comparison with #20 it allows to pass it  without removal of current default value `-Djava.security.egd=file:/dev/./urandom`. This looks to me as cleanest solution, if it covers this need, then no pb to merge it.

tobilarscheid:
Yes, I think #22 is a clean solution!

Godin:
Closed in preference to #22

Godin:
In fact `SONARQUBE_WEB_JVM_OPTS` (#22) doesn't work for `sonar.web.*`, however without any modifications, i.e. even for older images (e.g. 5.3), you can do
```
docker run -p 9000:9000 -it sonarqube:5.3 -Dsonar.web.context=/sq
```
arguments specified for `docker run` are passed to `jvm`. Thus, is there any benefits in addition of environment variable to specify the same?

IrishkA13:
Hi! 
I am trying to change 'sonar.web.context' property within Dockerfile (as ENV property) or docker-compose.yml (as environment property):

`ENV sonar.web.context /sonarqube`
or
`ENV JAVA_OPTS -Dsonar.web.context=/sonarqube`

Is there  a way to make it work? 
Or do I do somethin wrong?

Godin:
@IrishkA13 there is no such variables `sonar.web.context` and `JAVA_OPTS`, so that why it doesn't work. As explained in https://github.com/SonarSource/docker-sonarqube/issues/17#issuecomment-216856248 just above your comment, you can do:
```
[...]

PR #22, opened 2016-03-10
Should be possible to specify additional options for Web JVM

Godin:


StFS:
I ended up here after following https://github.com/SonarSource/docker-sonarqube/issues/17.

Unfortunately, specifying `sonar.web.context` by exporting `SONARQUBE_WEB_JVM_OPTS` in my `docker run` command does not seem to work.

I have an `/etc/systemd/system/sonarqube.service` file that looks like this:

```
[Unit]
Description=SonarQube service
Requires=docker.service postgres_sonarqube.service
After=docker.service postgres_sonarqube.service

[Service]
Restart=always
ExecStart=/usr/bin/docker run --name %n -p 9000:9000 \
          --link postgres_sonarqube.service:db \
          -e SONARQUBE_JDBC_USERNAME=sonarqube \
          -e SONARQUBE_JDBC_PASSWORD=sonarqube_password \
          -e SONARQUBE_JDBC_URL=jdbc:postgresql://db/sonarqube \
          -e SONARQUBE_WEB_JVM_OPTS=-Dsonar.web.context=/foobar \
          -v /mnt/data01/sonarqube/data:/opt/sonarqube/data \
          -v /mnt/data01/sonarqube/extensions:/opt/sonarqube/extensions \
          sonarqube:6.2
ExecStop=/usr/bin/docker stop %n
ExecStopPost=/usr/bin/docker rm -f %n

[Install]
WantedBy=multi-user.target
```

When I run this I get a sonarqube runnign on port 9000, but not under the `/foobar` context (it only works on the root context).

However, if I change the `sonar.properties` file in the container to have the `sonar.web.context` value set there, sonarqube works fine on the `/foobar` context (and not on the root context as expected).

Has anybody actually tested this and seen it work?

PR #24, opened 2016-04-08
Update SonarQube LTS to 4.5.7

Godin:


Godin:
This also requires merge of docker-library/official-images#1616

PR #26, opened 2016-04-19
reduce image size (alpine+less layers)

typekpb:
Reducing img sizes using alpinelinux base image. Moreover, decreasing layers count.

For comparison of image sizes:

	REPOSITORY            TAG  IMAGE ID            CREATED             SIZE
	sonarqube-orig        5.4  50e601571dc8        22 minutes ago      952.1 MB
	sonarqube             5.4  7ad728a66385        26 minutes ago      292.7 MB

where sonarqube-orig is current situation and sonarqube is the alpinelinux way.

### DIFF
--- a/4.5.7/Dockerfile
+++ b/4.5.7/Dockerfile
@@ -34,7 +34,7 @@ RUN apk --update upgrade \
 	&& rm sonarqube.zip* \
 	&& rm -rf $SONARQUBE_HOME/bin/* \
 	# clean apk cache
-	rm -rf /var/cache/apk/*
+	&& rm -rf /var/cache/apk/*
 
 VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions"]
 
--- a/4.5.7/Dockerfile
+++ b/4.5.7/Dockerfile
@@ -1,8 +1,14 @@
-FROM java:openjdk-8u45-jdk
+FROM java:openjdk-8u77-jdk-alpine
 
 MAINTAINER David Gageot <david.gageot@sonarsource.com>
 
-ENV SONARQUBE_HOME /opt/sonarqube
+ENV SONARQUBE_HOME=/opt/sonarqube \
+	# Database configuration
+	# Defaults to using H2
+	SONARQUBE_JDBC_USERNAME=sonar \
+	SONARQUBE_JDBC_PASSWORD=sonar \
+	SONARQUBE_JDBC_URL=jdbc:h2:tcp://localhost:9092/sonar \
+	SONAR_VERSION=4.5.7
 
 # Http port
 EXPOSE 9000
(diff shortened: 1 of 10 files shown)


Godin:
@typekpb first of all - thank you for your contribution! Reduction of size look really awesome :yum: :thumbsup:

I've no experience with Alpine, so wondering how all components of the SonarQube stack (JVM, ElasticSearch, SQ itself) behave on Alpine Linux - quoting https://hub.docker.com/_/java/ :
> certain software might run into issues

As an example - seems that official JRuby Docker image run into some troubles while switching to Alpine and this is quite interesting fact, because SQ uses JRuby. Also looking on official library - seems that not that much Java-based images started to use Alpine (I've checked some popular such as JRuby, Tomcat, Jenkins).

What do you think if we'll add Alpine image alongside with current ones to mitigate risks as it was done for example for Clojure? I.e. for current tags `lts` -> `4.5.7` and `latest` -> `5.4` at https://hub.docker.com/_/sonarqube/ we will perform only removal of unnecessary layers, but they will stay on Debian and will add `5.4-alpine` and `4.5.7-alpine`. Complete switch can be considered later.

Also we publish images via updates to https://github.com/docker-library/official-images/blob/master/library/sonarqube only for LTS and latest versions of SonarQube. And let's say that other intermediate versions preserved here only due/for historical reasons. So that no need to provide Alpine variant for them.

There also might be a nice side effect (to be checked), if switch to Alpine can help to overcome difficulties encountered in #18

4.5.7/Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN apk --update upgrade \
 	&& rm sonarqube.zip* \
 	&& rm -rf $SONARQUBE_HOME/bin/* \
 	# clean apk cache
-	rm -rf /var/cache/apk/*
```
Godin: Hmm, interesting why build in Travis passed successfully with this.

Godin:
Amount of layers reduced in e7a075a946d8b92dfa8fe707407d5ccd994c0d46

typekpb:
@Godin
I can't really comment on possible issues, but providing separate alpine
based docker images sounds like the way to go.

Regarding forgotten `&&`, no clue why that passed.

Reduced layer count seems safe so +1 for considering that.
On Apr 23, 2016 15:12, "Evgeny Mandrikov" <notifications@github.com> wrote:

Amount of layers reduced in e7a075a
<https://github.com/SonarSource/docker-sonarqube/commit/e7a075a946d8b92dfa8fe707407d5ccd994c0d46>

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/SonarSource/docker-sonarqube/pull/26#issuecomment-213738696>

Godin:
Ok, images based on Alpine added in 1d696886185a5f8aa345b3040153ba92769c34e4

Both changes will be available in official library after merge of https://github.com/docker-library/official-images/pull/1659

PR #28, opened 2016-05-02
SonarQube 5.5

Godin:


Godin:
This also requires merge of docker-library/official-images#1702

PR #29, opened 2016-05-15
changing plugin data volume to use base container

mpeter:
If we use the sonarqube image for the plugins data volume, it's going to default to use the hsql database, as well as run an actual instance of sonarqube, which we don't want in this recipe. This update ensures that we use the base jdk container instead.

### DIFF
--- a/recipes.md
+++ b/recipes.md
@@ -29,7 +29,7 @@ datadb:
   command: /bin/true
 
 plugins:
-  image: sonarqube
+  image: java:openjdk-8u45-jdk
   volumes:
    - /opt/sonarqube/extensions
    - /opt/sonarqube/lib/bundled-plugins


Godin:
@mpeter I think would be better to avoid name of base image in recipe in general, because otherwise we'll need to not forget to update it in case of change of base image.

Also `*-alpine` uses another base image - `java:8-alpine`.

But most importantly - execution of
```
docker run sonarqube /bin/true
```
doesn't start SonarQube and terminates, because there is check in `run.sh` - see https://github.com/SonarSource/docker-sonarqube/blob/445337a0bbe87afd8ba745ba4bcabf75d3bf002e/4.5.7/run.sh#L5
So I'm not sure that there is an issue with recipe.

mpeter:
Up to you -- I had a heck of a time getting this to work in my environment once you assume that the postgresql server was not the default one.

Godin:
@mpeter No problem to fix this in one way or another, if problem can be confirmed. So I'm wondering how to reproduce this. Could you please provide some logs, which shows that "plugins" container actually starts SonarQube.?

Godin:
Closing this in absence of response on https://github.com/SonarSource/docker-sonarqube/pull/29#issuecomment-219530442

PR #36, opened 2016-07-14
move run script COPY to allow for better volume mounting

Skilgarriff:
What this pull request does is simply move where the run script is COPYed into the container. I was attempting to run this container by simply mounting the volume /opt/sonarqube instead of all of the smaller suggested volumes. This is because on a kubernetes cluster it is infeasible to spin up seperate architecture for each small volume if we want persistance, thus I thought it would make more sense to simple mount one volume. 

The issue is in the order of events. A volume mount wipes anything that is already there thus:
COPY to /opt/sonarqube/bin
Volume is mounted -> /opt/sonarqube is wiped
ENTRYPOINT is attempted to be run -> run script no longer exists. 

By moving the run script outside of the /opt/sonarqube directory then it can survive a broader volume mount.

### DIFF
--- a/5.6-alpine/Dockerfile
+++ b/5.6-alpine/Dockerfile
@@ -35,5 +35,5 @@ RUN set -x \
 VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions"]
 
 WORKDIR $SONARQUBE_HOME
-COPY run.sh $SONARQUBE_HOME/bin/
-ENTRYPOINT ["./bin/run.sh"]
+COPY run.sh /
+ENTRYPOINT ["/run.sh"]
--- a/5.6-alpine/run.sh
+++ b/5.6-alpine/run.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+cd /opt/sonarqube
+
 set -e
 
 if [ "${1:0:1}" != '-' ]; then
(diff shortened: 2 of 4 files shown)


Skilgarriff:
Is there anything I can do to help this along? I can't build SonarQube on Kubernetes in a reasonable way until this gets merged. Currently to make Sonar persistent we would need 4 volumes, and thus 4 EBS volumes on AWS. With this merge, we could only use 1 volume.

Godin:
Hi @Skilgarriff 
> Is there anything I can do to help this along?

be a little bit more patient? yesterday was a day off in France for me and for many other SonarSource developers 😉 

Even if I didn't fully understand the underlying problem, I'm ready to merge this, but wondering:

1. will it be ok for you, if script will be in directory `/root` instead of `/` and `/opt/sonarqube` ? This looks more logical for me. If so, then please update PR.
2. maybe there is someone, who depends on current location of this script? Not sure what is better - break their scripts or take some debt here by maintaining compatibility.

Also, merge of this won't magically result in update for Docker Hub - publishing requires a separate PR from us for https://github.com/docker-library/official-images From here I'm wondering why you can't simply use your own local Dockerfile with desired changes or publish it to Hub under your account to unblock your work with Kubernetes and AWS while waiting for merge of this PR?

Godin:
We are going to roll out PR with SonarQube 5.6.1 for Docker Hub, and this is a good opportunity to include this change, if there are still demand. So @Skilgarriff is there something I can do to help you to address raised questions to finalize this PR?

Godin:
No response - no changes, sorry.

PR #42, opened 2016-08-09
Add the ability to set arguments to Java when running Sonar.

pwagland:
I need to be able to give sonarqube more memory so that it doesn't get an OOM trying to process our tree, but the standard recipes don't allow for that. This pull request is to add that capability.

### DIFF
--- a/5.6.1-alpine/run.sh
+++ b/5.6.1-alpine/run.sh
@@ -6,8 +6,7 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
-  ${SONARQUBE_JAVA_ARGS} \
+exec java ${SONARQUBE_JAVA_ARGS} -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
   -Dsonar.jdbc.password="$SONARQUBE_JDBC_PASSWORD" \
--- a/5.6.1-alpine/run.sh
+++ b/5.6.1-alpine/run.sh
@@ -7,6 +7,7 @@ if [ "${1:0:1}" != '-' ]; then
 fi
 
 exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+  ${SONARQUBE_JAVA_ARGS} \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
   -Dsonar.jdbc.password="$SONARQUBE_JDBC_PASSWORD" \
--- a/6.0-alpine/run.sh
+++ b/6.0-alpine/run.sh
@@ -6,8 +6,7 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
-  ${SONARQUBE_JAVA_ARGS} \
+exec java ${SONARQUBE_JAVA_ARGS} -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
   -Dsonar.jdbc.password="$SONARQUBE_JDBC_PASSWORD" \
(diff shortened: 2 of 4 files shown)


pwagland:
Figured out how to test this locally… and did so, so now it really works instead of just me thinking that it should work…

Mea Culpa!

5.6.1-alpine/run.sh
```diff
@@ -7,6 +7,7 @@ if [ "${1:0:1}" != '-' ]; then
 fi
 
 exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+  ${SONARQUBE_JAVA_ARGS} \
```
Godin: This seems completely useless, because arguments specified for `docker run` as
```
docker run -p 9000:9000 -it sonarqube:5.6.1 argument
```
already will appended - see https://github.com/SonarSource/docker-sonarqube/blob/73534fd899d29eb1373ea073edbfe7a4074ede6e/5.6.1/run.sh#L15

6.0-alpine/run.sh
```diff
@@ -6,8 +6,7 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
-  ${SONARQUBE_JAVA_ARGS} \
+exec java ${SONARQUBE_JAVA_ARGS} -jar lib/sonar-application-$SONAR_VERSION.jar \
```
Godin: `sonar-application-$SONAR_VERSION.jar` is just a bootstrapper, which will launch other java processes, and so I have doubts that setting `-Xmx` here would affect those child processes

pwagland: I fear that you might be correct :-(

How should those be set, since I can't modify the config file at the moment?

Godin:
@pwagland Given the comments above, could you please provide usage scenario (commands, logs, ...) showing the difference - before this change and after?

6.0-alpine/run.sh
Godin: @pwagland my best guess (taking into account current time) to be verified - is that you can do it as:
```
docker run -p 9000:9000 -it sonarqube:5.6.1 -Dsonar.ce.javaOpts=-Xmx...
```
See https://github.com/SonarSource/sonarqube/blob/5.6.1/sonar-application/src/main/assembly/conf/sonar.properties#L137

pwagland:
Thanks for your hint!

So I need this change to be able to get the argument in from docker-compose. Docker-compose doesn't allow you to pass arguments through, but you can set environment variables.

So I have added:

```
    environment:
      - SONARQUBE_JAVA_ARGS=-Dsonar.ce.javaOpts=-Xmx2048m -Dsonar.web.javaOpts=-Xmx2048m
```

to my `docker-compose.yml`, and now I can upload my project results. My project has a reasonable sized report:

```
INFO: Analysis report generated in 37530ms, dir size=342 MB
INFO: Analysis reports compressed in 59431ms, zip size=115 MB
INFO: Analysis report uploaded in 4871ms
```

And without the above arguments, it doesn't allow the upload to succeed.

Godin:
@pwagland it is hard to believe that docker-compose doesn't allow to pass arguments - see https://docs.docker.com/compose/compose-file/#/command

I'm pretty sure that you already can do:
```
version: "2"

services:
  sonarqube:
    image: sonarqube:5.6.1
    command: -Dsonar.ce.javaOpts=-Xmx2048m -Dsonar.web.javaOpts=-Xmx2048m
```

Godin:
which for example results in `-Xmx2048m` being correctly passed:
```
sonarqube_1  | 2016.08.10 06:35:46 INFO  app[o.s.p.m.JavaProcessLauncher] Launch process[ce]: /usr/lib/jvm/java-1.8-openjdk/jre/bin/java -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Xmx2048m -Djava.io.tmpdir=/opt/sonarqube/temp -javaagent:/usr/lib/jvm/java-1.8-openjdk/jre/lib/management-agent.jar -cp ./lib/common/*:./lib/server/*:./lib/ce/*:/opt/sonarqube/lib/jdbc/h2/h2-1.3.176.jar org.sonar.ce.app.CeServer /opt/sonarqube/temp/sq-process7296139030484168998properties
```

pwagland:
Ah! You are right. Mea-culpa.

I had looked at the 'args' on that page, for passing args to the build process, and could not find anything equivalent for the 'normal run'.

Thanks, and sorry for the noise, this can be closed.

PR #45, opened 2016-08-24
expose conf and logs directory as docker volumes

andreyhsiao:


6.0/Dockerfile
```diff
@@ -37,9 +37,11 @@ RUN set -x \
     && unzip sonarqube.zip \
     && mv sonarqube-$SONAR_VERSION sonarqube \
     && rm sonarqube.zip* \
-    && rm -rf $SONARQUBE_HOME/bin/*
+    && rm -rf $SONARQUBE_HOME/bin/* \
+    && mkdir -p $SONARQUBE_HOME/samples/conf \
+    && mv $SONARQUBE_HOME/conf/* $SONARQUBE_HOME/samples/conf
```
Godin: @andreyhsiao I guess that this relocation here in order to pre-populate an empty volume, since volume hides existing content. I believed that in case of data-only containers good practice is to re-use image, so that data will be there. Thus best guess is that this is for the case of mount of a real file-system. Which I assume anyway not a good practice. So could you please elaborate a little bit more about this relocation?

andreyhsiao:
@Godin Sure, this is for mounting file-system volumes, which can ease modifying configs.
What will be the recommended way for exposing the conf folder, thanks.

Godin:
@andreyhsiao I think that initialization of empty folder can be removed and we can stick to a simple statement:
> If you mount host directory, then it must contain pre-existing configuration.
> Initial configuration can be obtained by manual copy from an image.

Note that this allows to avoid bad surprises for the consumers, who expect that `/opt/sonarqube/conf` exists as it is now.

Will this work for you? If so, then please update PR accordingly. And in order to finalize this PR the same changes should be applied for other versions, not just for `6.0`.

andreyhsiao:
@Godin Agreed, the changes should be backward compatible.

I have updated the changes, please help review.

5.6.1/run.sh
```diff
@@ -2,6 +2,10 @@
 
 set -e
 
+if [ ! $(ls -A "$SONARQUBE_HOME"/conf) ]; then
```
Godin: @andreyhsiao First - could you please finally explicitly answer on question posed in https://github.com/SonarSource/docker-sonarqube/pull/45#issuecomment-242339735 : why you want to automate copying?

Second - check of emptiness of directory is flawed:
```
$ docker build -t test 5.6.1
$ docker run --rm -it test
./bin/run.sh: line 5: [: sonar.properties: unary operator expected

$ docker build -t test 5.6.1-alpine
$ docker run --rm -it test
sh: wrapper.conf: unknown operand
```
I believe that correct check is:
```
if [ -z "$(ls -A "$SONARQUBE_HOME"/conf)" ]; then
```

andreyhsiao:
@Godin The reason is for mounting file-system volumes, which will hide the existing conf content.

If it doesn't seem as necessary, I will close this PR at this point.

Thank you

Godin:
@andreyhsiao Question was not about mount of local file-system, which obviously hides existing content in a container, but why you want pre-initialization? Let me demonstrate.

Let's create config in local directory and mount it as a volume:
```
$ echo "sonar.web.javaOpts=demo" > sonar.properties`
$ docker run --rm -it -v $PWD:/opt/sonarqube/conf sonarqube:5.6.1-alpine
Error: Could not find or load main class demo
```
startup failed, demonstrating that config is taken into account.

Mount hides files, but empty directory is not a problem for startup:
```
$ docker run --rm -it sonarqube:5.6.1-alpine ls -A /opt/sonarqube/conf
sonar.properties  wrapper.conf
$ ls -lA
total 0
$ docker run --rm -it -v $PWD:/opt/sonarqube/conf sonarqube:5.6.1-alpine ls -lA /opt/sonarqube/conf
total 0
$ docker run --rm -it -v $PWD:/opt/sonarqube/conf sonarqube:5.6.1-alpine
```

Finally if you want to obtain original config from container, you can do:
```
$ ls -lA
total 0
$ docker run --name conf sonarqube:5.6.1-alpine true
$ docker cp conf:/opt/sonarqube/conf/. .
$ docker rm conf
$ ls -A
sonar.properties  wrapper.conf
```

The question that you have not answered so far:
why
"Initial configuration can be obtained by manual copy from an image." (  https://github.com/SonarSource/docker-sonarqube/pull/45#issuecomment-242339735 )
is not acceptable for you and why you want to automate this?

Godin:
Superseded by #56

PR #53, opened 2016-10-22
Use full path on entrypoint

luisgros-fixeads:


5.6.3/Dockerfile
```diff
@@ -34,4 +34,4 @@ VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions"]
 
 WORKDIR $SONARQUBE_HOME
 COPY run.sh $SONARQUBE_HOME/bin/
-ENTRYPOINT ["./bin/run.sh"]
+ENTRYPOINT ["$SONARQUBE_HOME/bin/run.sh"]
```
Godin: This doesn't work for me:
```
$ docker --version
Docker version 1.12.0, build 8eab29e
$ cd 5.6.3
$ docker build -t test .
$ docker run -it --rm test
docker: Error response from daemon: oci runtime error: exec: "$SONARQUBE_HOME/bin/run.sh": stat $SONARQUBE_HOME/bin/run.sh: no such file or directory.
```

Godin: @luisgros-fixeads do you plan to say something about failure above?

Godin:
No response - no change.

Issue #34, opened 2016-06-28
Give access to sonar.properties


osechet:
Some plugins reads their configuration from the `sonar.properties` file in the `conf` directory. It could be nice to have an access to that file from the host (i.e. with a volume).

andreas-lundgren-storytel:
Hi!

The mapping to a persisted volume is done at line 13 in recipes.md, "sonarqube_conf:/opt/sonarqube/conf"

If you want to map the conf folder to a specific folder on the host, you can change from a named volume to a path mapped volume according to https://docs.docker.com/compose/compose-file/#volumes-volume-driver.

osechet:
Sorry, my mistake. I made a typo when writing the volume path on the command line and I couldn't make it work. It's now working well.

LorenzoBettini:
@andreas-lundgren-storytel but since the conf directory is not exposed as a volume, mounting locally will not generate any default configuration file, the same for bundled-plugins, it will be created empty

LorenzoBettini:
Sorry, my comment is completely wrong. Locally mounted directories have nothing to do with the volumes defined in the docker compose file. I made confusion.

Issue #16, opened 2016-01-12
document volumes


magnus919:
The current Dockerfile defines two different volumes. I saw a question on Docker Hub about whether or not there are volumes on this image, so it's probably worth having some doc around that.

Godin:
Docs for Docker Hub are located in https://github.com/docker-library/docs/tree/master/sonarqube Feel free to open PR and suggest change.

dhoer:
Can those volumes be removed from the Dockerfile and replaced with docker and docker-compose examples on how to mount using volumes in the readme.  Having volumes defined in the Dockerfile makes a trivial task of adding plugins complex.

Godin:
@dhoer hm, your proposal conflicts with what other people propose - #44 , #56 
As a maintainer of this repository I'm fine to accept either of solutions, given good argumentation for it.
So feel free to elaborate more here about Docker best practices about this subject and provide patch.

LorenzoBettini:
I'm kind of a newbie with Docker, but how would volumes defined in the Dockerfile get in your way when using mounted directories or when using docker-compose? As far as I know, the ones defined in the docker-compose would still work and if you mount a local directory that would take the precedence... or am I missing something?

Godin:
See some advices for SonarQube image from @tianon about volumes in https://github.com/docker-library/official-images/pull/2437#issuecomment-266573937

pierre-guillot-sonarsource:
we updated the doc month ago

PR #56, opened 2016-12-03
expose conf and logs directory as docker volume

### LINKED ISSUE #44: expose conf and logs directory as docker volume
exposing conf and logs directory as docker volume

LorenzoBettini:
Closes #44

### DIFF
--- a/5.6.3-alpine/Dockerfile
+++ b/5.6.3-alpine/Dockerfile
@@ -32,7 +32,7 @@ RUN set -x \
     && rm sonarqube.zip* \
     && rm -rf $SONARQUBE_HOME/bin/*
 
-VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions"]
+VOLUME ["$SONARQUBE_HOME/data", "$SONARQUBE_HOME/extensions", "$SONARQUBE_HOME/conf", "$SONARQUBE_HOME/logs"]
 
 WORKDIR $SONARQUBE_HOME
 COPY run.sh $SONARQUBE_HOME/bin/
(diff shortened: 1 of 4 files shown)


Godin:
@LorenzoBettini  first of all excuse me that it took so long to take a look on this - was on vacation and traveling.

Unfortunately detailed information is not provided along with your change as well as in #44. So based on your https://github.com/SonarSource/docker-sonarqube/issues/34#issuecomment-264516923 I tried to guess that this is about command such as:
```
docker run -v /tmp/sq/conf:/opt/sonarqube/conf sonarqube sh
```
and you expect that file `/tmp/sq/conf/sonar.properties` will be created, but this is not the case even after this change - quoting https://docs.docker.com/engine/tutorials/dockervolumes/ :
> Volumes are initialized when a container is created. If the container’s base image contains data at the specified mount point, that existing data is copied into the new volume upon volume initialization. (**Note that this does not apply when mounting a host directory.**)

So could you please elaborate more about rationale for this change?

LorenzoBettini:
@Godin of course my comment https://github.com/SonarSource/docker-sonarqube/issues/34#issuecomment-264516923 is completely wrong, as you said, and I forgot to update it.  This Pr is just for #44 which I seem to understand would still be useful, so that conf and logs are persisted in volumes that can be later inspected or modified. So it has nothing to do with mounted host directories.

Godin:
@LorenzoBettini thank you for your contribution! :thumbsup:

Godin:
@LorenzoBettini after merge of this PR came across https://github.com/SonarSource/docker-sonarqube/issues/16#issuecomment-256426270 Probably I lack knowledge of Docker best practices to understand all this mess around volumes, so could you please participate in mentioned thread? Thanks in advance.

Godin:
Reverted in https://github.com/SonarSource/docker-sonarqube/commit/80366e3419d698b4bba4447f153418ef64b3b705

PR #66, opened 2017-01-10
it tried to find /lib/... in the current /bin/ folder and failed to e…

ovidiugrebenisan:
Executing run.sh fails because it is placed in SONARQUBE_HOME/bin and tries to find lib/sonar-application-$SONAR_VERSION.jar inside that folder.


Godin:
Closing this in absence of response on https://github.com/SonarSource/docker-sonarqube/pull/66#pullrequestreview-16703130

PR #69, opened 2017-01-14
Added variable for jvm options

cmroche:
Gives the ability to better manage memory pool params (and other java opts) through the environment variables used to launch the docker instance. 512M was set as a reasonable local service size for my purposes, but possibly should be bigger, perhaps 1GiB.

### DIFF
--- a/5.5/run.sh
+++ b/5.5/run.sh
@@ -6,7 +6,7 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+exec java $SONARQUBE_JAVA_OPTS -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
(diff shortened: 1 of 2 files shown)


cmroche:
I'll need to update locally first due to conflict.

5.5/run.sh
```diff
@@ -6,7 +6,7 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+exec java $SONARQUBE_JAVA_OPTS -jar lib/sonar-application-$SONAR_VERSION.jar \
```
Godin: @cmroche note that `sonar-application.jar` just starts and monitors other java processes (such as web server, elastic search), so setting `-Xmx` has no effect on them. Settings for those processes can be placed into `conf/sonar.properties` (see [`sonar.web.javaOpts` for example](https://github.com/SonarSource/sonarqube/blob/5.5/sonar-application/src/main/assembly/conf/sonar.properties#L93)). Or specified as parameters to `sonar-application.jar` and so already supported by this image without need to introduce any additional environment variables:
```
docker run -d --name sonarqube -p 9000:9000 sonarqube -Dsonar.web.javaOpts=-Xmx512m
```

PR #84, opened 2017-05-31
Remove deprecated MAINTAINER instruction

ben-st:
maintainer is deprecated in release: v1.13.0
https://docs.docker.com/engine/deprecated/#repositoryshortid-image-references
I would like to make the change to label instead


Godin:
Thank you for pointing on deprecation notice. Actually current value is outdated :wink: , and looking on some other images in https://github.com/docker-library/official-images such as Jenkins, Jetty, Gradle, and for the reasons outlined in https://github.com/moby/moby/pull/25466#issuecomment-239488477 , IMO we can simply remove it. I updated your branch to do removal.

Thank you for your contribution :thumbsup:

Issue #93, opened 2017-06-15
Provide standard way to install plugins


tobilarscheid:
Hi everyone,

I think this image should have a standard way to install required sonar plugins, maybe by passing them in as a list or something like this.

Regards,

Tobi

jkblume:
this would be nice

madnight:
Sonarqube should provide a plugin manager. For instance, both zsh and vim have many different plugin managers available, e.g. https://github.com/zplug/zplug for zsh or https://github.com/junegunn/vim-plug for vim.

tobilarscheid:
Hi @jkblume @madnight please have a look at #94 and let me know what you think!

Godin:
Closing this for the reasons described in #94

PR #94, opened 2017-06-27
Provide standard way to install plugins

tobilarscheid:
My proposal for solving #93 

To install plugins from the sonarqube plugin library, you pass all desired plugins as a list in an environment variable, e.g. like so:

`docker run -e SONARQUBE_PLUGIN_LIST=sonar-github-plugin:1.4.1.822,sonar-flex-plugin:2.3 library/sonarqube`

Already installed plugins are skipped automatically and any existing version (if not the desired one) is deleted to avoid conflicts.

I will also provide a PR for updating the documentation in [docker-library/docs/sonarqube](https://github.com/docker-library/docs/tree/master/sonarqube) and reference it from here.

I am very interested in your feedback!

Regards,

Tobi

### DIFF
--- a/6.4-alpine/run.sh
+++ b/6.4-alpine/run.sh
@@ -6,6 +6,28 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
+#Install additional plugins
+SONAR_PLUGIN_DIR="/opt/sonarqube/extensions/plugins"
+NUMBER_OF_PLUGINS=$(($((`echo $SONARQUBE_PLUGIN_LIST | sed 's/[^,]//g' | wc -c` - 1 ))+1))
+for i in $(seq 1 $NUMBER_OF_PLUGINS);
+do
+    #Extract plugin name and version
+    pluginName=$(echo "$SONARQUBE_PLUGIN_LIST" | cut -d, -f$i | cut -d: -f1)
+    pluginVersion=$(echo "$SONARQUBE_PLUGIN_LIST" | cut -d, -f$i | cut -d: -f2)
+    #Skip if desired plugin version is already installed
+    if [ -e $SONAR_PLUGIN_DIR/$pluginName-$pluginVersion.jar ]
+    then
+        echo "skipping $pluginName as it is already installed in version $pluginVersion"
+        continue
+    fi
+    #Remove any existing artifacts with wrong version
+    rm -f $SONAR_PLUGIN_DIR/$pluginName*
+    #Download desired plugin
+    echo "Installing $pluginName-$pluginVersion"
+    wget -O $SONAR_PLUGIN_DIR/$pluginName-$pluginVersion.jar --no-verbose "https://sonarsource.bintray.com/Distribution/$pluginName/$pluginName-$pluginVersion.jar"
+done
+
+#Run sonar
 exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
--- a/6.4/run.sh
+++ b/6.4/run.sh
@@ -6,6 +6,28 @@ if [ "${1:0:1}" != '-' ]; then
   exec "$@"
 fi
 
+#Install additional plugins
+IFS=',' read -r -a plugins <<< "$SONARQUBE_PLUGIN_LIST"
+SONAR_PLUGIN_DIR="/opt/sonarqube/extensions/plugins"
+for plugin in "${plugins[@]}"
+do
+    #Extract plugin name and version
+    IFS=':' read -r -a pluginNameAndVersion <<< "$plugin"
+    pluginName=${pluginNameAndVersion[0]}
+    pluginVersion=${pluginNameAndVersion[1]}
+    #Skip if desired plugin version is already installed
+    if [ -e $SONAR_PLUGIN_DIR/$pluginName-$pluginVersion.jar ]
+    then
+        echo "skipping $pluginName as it is already installed in version $pluginVersion"
+        continue
+    fi
+    #Remove any existing artifacts with wrong version
+    rm -f $SONAR_PLUGIN_DIR/$pluginName*
+    #Download desired plugin
+    echo "Installing $pluginName-$pluginVersion"
+    curl -o $SONAR_PLUGIN_DIR/$pluginName-$pluginVersion.jar -fSL "https://sonarsource.bintray.com/Distribution/$pluginName/$pluginName-$pluginVersion.jar"
+done
+
 exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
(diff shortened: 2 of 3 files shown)


tobilarscheid:
The corresponding documentation PR is open: https://github.com/docker-library/docs/pull/952

Godin:
@tobilarscheid first of all thank you for the proposal!

However even without this change already possible to install plugins using a standard Docker way of extending images as following for example:
```
FROM sonarqube:6.3.1

RUN curl -o /opt/sonarqube/extensions/plugins/ https://sonarsource.bintray.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-3.1.0.5111.jar
```
Or by copying them from local filesystem or by mounting volume.
IMO proposed change is unnecessary non-flexible overcomplication, which increases maintenance costs (e.g. in case of change of URLs) compared to the ways described above.
What IMO will be better is to document such standard options for end users.

Thank you for your understanding.

madnight:
The plugin installation should be versioned and it should be updatable and it should be docker independent. All these condiditions are currently unmet and the proposal does not address these basic plugin manager features.

tobilarscheid:
Hi @Godin,

thanks for your quick feedback. I am aware of the way you described, actually I even used it in several projects.

I do however feel it has one big disadvantage: You have to actually build a new container. Building a new container also means you need a delivery infrastructure for this image, i.e. you need a docker registry if you want to build this container once and then run it on multiple hosts. (Think kubernetes for example)

I am aware that both ways (my way and extending image) have their specific advantages and disadvantages and I do think they could happily co-exist. My way might be more helpful for quick try-out, the other way is for more static scenarios.

I understand your argument regarding maintenance of changing URLs, I do however hold against it that with the extend-image way we put this maintenance burden on all users of the image - instead of just managing it central.

tobilarscheid:
Hi @madnight,

thanks for your feedback. You are absolutely right, but the issue you describe is not related to docker-sonarqube and should probably instead be addressed at the [Sonar Source Issue Tracker](https://jira.sonarsource.com/browse/SONAR)

Godin:
@tobilarscheid being the only for the time being maintainer of this repository with a limited time resources and given the low contributions activity, I'd prefer to keep things simple.

@madnight "plugin manager" is out of scope of goals of this repository that hosts solely sources to build Docker image containing SonarQube.

tobilarscheid:
@Godin I promise to provide PRs for any URL change that will happen :-)

Godin:
@tobilarscheid 
> Building a new container also means you need a delivery infrastructure for this image

IMO infrastructure is anyway required for effective usage of containers. Local build of an image doesn't seem for me to be different from case when plugins are downloaded by image. Or even simpler - mount local directory with plugin. BTW you can store your images at https://hub.docker.com/ isn't it? 😉 

> I do think they could happily co-exist

on this let me refer to Occam's razor 😉 

> I promise to provide PRs for any URL change that will happen

This is not just about URL change, but also about questions and answers that appear from time to time, about basic tests done when updating Dockerfile, etc. However nice commitment, but you never know what tomorrow will bring, and only time can show if commitment fulfilled in mid/long-term. In terms of your involvement/contributions as a way to show commitment - why not start from documenting really standard ways to extend image? and there are btw some tickets marked as "help wanted" :wink:

martinsson:
@Godin the proposed solution
  
  FROM sonarqube:6.3.1
  RUN curl -o /opt/sonarqube/extensions/plugins/ https://sonarsource.bintray.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-3.1.0.5111.jar

doesn't work for me. On startup using docker-compose the plugins directory is replaced by the contents of the volume.

I don't understand why this doesn't happen to you. I'm not sure what solution would be best. Extending the volume sonarqube_extensions?

tobilarscheid:
@martinsson I think you are mixing two solutions:
- Either you mount the volume (and then put the plugin jars in the mapped directory on the host)
- or you build your own image where you `curl` the plugin jars (like in the snippet you posted)

Mixing both solutions will - as you correctly identified - not work properly.

martinsson:
Yes and, I was looking to extend the existing volume, not replacing all plugins.
Why is the docker compose using a volume for extensions anyway? Volumes in my understanding are immutable and not extensible. 

I'd like to have an easy way of just adding my plugins. One way would be if the Dockerfile (official one) loaded all bundled plugins. That would also act as documentation, and pretty much solve #16 

My docker-fu is limited though, so I'm assuming there are better solutions

tobilarscheid:
Sorry, I can't speak for the `docker-compose`, @Godin knows more :-)

Godin:
@martinsson 

> the proposed solution
> ```
> FROM sonarqube:6.3.1
> RUN curl -o /opt/sonarqube/extensions/plugins/ https://sonarsource.bintray.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-3.1.0.5111.jar
> ```
> doesn't work for me

This probably doesn't work because as of today `curl` gets no bytes from bintray for some unknown to me reasons. Leaving this aside as well as `docker-compose` and replacing `curl` by `wget`:

```
$ docker run -it --rm sonarqube:6.3.1 ls /opt/sonarqube/extensions/plugins
README.txt

$ cat << END > Dockerfile

END

$ docker build -t test .

$ docker run -it --rm test ls /opt/sonarqube/extensions/plugins
README.txt  sonar-javascript-plugin-3.1.0.5111.jar
```

as you can see - third command shows that plugin was added to the image and during execution of this image it will be used.

> On startup using docker-compose the plugins directory is replaced by the contents of the volume.

Now if we speak about `docker-compose`, then how your `docker-compose.yml` looks like?

If I use following `docker-compose.yml` :

```
version: "2"

services:
  sonarqube:
    build: .
```

together with `Dockerfile` created on previous step, then `docker-compose up` will start SonarQube with added version of JavaScript plugin (`3.1.0.5111`).

> I was looking to **extend** the existing volume [...] Volumes in my understanding are immutable and **not extensible**.

Contradictory statements, don't you think so? And again there is lack of details / explanations about what and how you're trying to do, especially that there is a way too many ways of doing things using docker.

> One way would be if the Dockerfile (official one) loaded all bundled plugins.

In fact during startup SonarQube always automatically installs all "bundled" plugins (particular versions of plugins such as Java, JavaScript, etc; coming with particular version of SonarQube) from directory `lib/bundled-plugins` into directory `extensions/plugins` - this mechanism is independent from docker image.

Godin:
@tobilarscheid as this thread shows - this change will bring even more complexity with understanding of how things work. So let me refer to Occam's razor again and close this PR. Thank you for your understanding.

tobilarscheid:
@Godin, I respect your decision. I just wanted to leave one last comment telling you that I personally think that my solution could resolve a lot of this confusion. Obviously people are not comfortable with low level `Dockerfile` and volume handling when all they want to do is add some plugins.
But once again, I know that maintaining such a repo is a difficult job and fully respect your decision to leave everything as is.

Regards,

Tobi

PR #96, opened 2017-07-06
Improve layers, streamline docker cache

NonCreature0714:
While possible, multi-line ENV is not necessary or recommended by Docker, and preserving each ENV variable allows docker to check each layer containing and ENV for changes, instead of treating any change of ENVs as a layer change. Copy 'run.sh' at a lower layer to enhance Docker cache as it is faster to execute than 'RUN apt-get' and establishes a common layer, lower.

### DIFF
--- a/6.4-alpine/Dockerfile
+++ b/6.4-alpine/Dockerfile
@@ -1,15 +1,14 @@
 FROM openjdk:8-alpine
 
-ENV SONAR_VERSION=6.4 \
-    SONARQUBE_HOME=/opt/sonarqube \
-    # Database configuration
-    # Defaults to using H2
-    SONARQUBE_JDBC_USERNAME=sonar \
-    SONARQUBE_JDBC_PASSWORD=sonar \
-    SONARQUBE_JDBC_URL=
+ENV SONAR_VERSION=6.4
+ENV SONARQUBE_HOME=/opt/sonarqube
+# Database configuration
+# Defaults to using H2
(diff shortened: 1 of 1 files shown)


NonCreature0714:
Check CI check failed because of "/bin/sh: 1: gpg: not found". This is not something I altered.

NonCreature0714:
@Godin Sorry for a delayed response. This is based on a PR I made to [Docker Library's httpd](https://github.com/docker-library/httpd), where I consolidated ENV into one layer. [yosifkit ](https://github.com/yosifkit) Docker's official image maintainer, closed my PR, [stating that putting all ENVs into one layer is unnecessary](https://github.com/docker-library/httpd/pull/58).

Godin:
@NonCreature0714 thanks for explanation about `ENV`. What about other changes?

NonCreature0714:
@Godin  Thanks for asking. It's my understanding that layers more prone to changes/overrides should be placed later in a Dockerfile, to better optimize the docker-cache so there aren't so many low-level layer variations. Thus I moved the COPY early because it's less likely to change, and moved EXPOSE later because it is something more frequently overridden.

NonCreature0714:
I'm going to close this PR based on a response to a question [I asked on StackOverflow](https://stackoverflow.com/questions/45291314/dockerfile-is-my-placement-of-expose-correct) -- Looks like I was on the wrong track, but such is learning.

Godin:
@NonCreature0714 Thanks for investigations and for sharing the result! :thumbsup:

PR #104, opened 2017-08-28
Need depends_on

philipz:
If this docker-compose.yml doesn't have "depends_on", sonarqube container will down.

### DIFF
--- a/recipes.md
+++ b/recipes.md
@@ -14,6 +14,8 @@ version: "2"
 
 services:
   sonarqube:
+    depends_on:
+      - db
     image: sonarqube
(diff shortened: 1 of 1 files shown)


philipz:
Yes, if no "depends_on", after "docker-compose up" the sonarqube container will run quicker than postgres container. Add "depends_on" could fix this problem. So, it's up to you.

simonbrandhof:
I agree with @Godin. The option `depends_on` does not ensure that the database is operational. As the check to be implemented depends on the DB vendor, I suggest to do nothing in the documented recipe.

PR #109, opened 2017-09-26
SonarQube 5.6.7

Godin:


Godin:
This also requires merge of https://github.com/docker-library/official-images/pull/3496

tianon:
This needed to update `SONAR_VERSION` as well, right?

Godin:
@tianon ouch, you're right, I was too fast

Godin:
@tianon fixed in 65090c41e5e8157e656e6a7fd3853f5e09f6eae6

PR #111, opened 2017-09-28
Manage sonarqube data volumes at the highest level

Theaxiom:
It is best practice to manage the docker volumes at the highest level for container data. It makes it much easier to manage as well as ensure scalability for the future when additional directories are added.

### DIFF
--- a/recipes.md
+++ b/recipes.md
@@ -22,10 +22,7 @@ services:
     environment:
       - SONARQUBE_JDBC_URL=jdbc:postgresql://db:5432/sonar
     volumes:
-      - sonarqube_conf:/opt/sonarqube/conf
-      - sonarqube_data:/opt/sonarqube/data
-      - sonarqube_extensions:/opt/sonarqube/extensions
-      - sonarqube_bundled-plugins:/opt/sonarqube/lib/bundled-plugins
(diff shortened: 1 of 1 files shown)


Theaxiom:
I see now why they are setup that way, it is extracting those directories from the root:
```
root@localhost:/opt/sonarqube# ls -la
total 12
drwxr-xr-x.  1 root root   30 Aug  1 12:10 .
drwxr-xr-x.  1 root root   23 Sep 14 07:45 ..
-rw-r--r--.  2 root root 7651 Aug  1 12:00 COPYING
drwxr-xr-x.  2 root root   20 Sep 14 07:45 bin
drwxr-xr-x.  2 root root    6 Sep 28 18:49 conf
drwxr-xr-x.  4 root root   27 Sep 28 18:49 data
drwxr-xr-x.  3 root root   23 Sep 28 18:49 extensions
drwxr-xr-x. 10 root root  152 Sep 28 18:49 lib
drwxr-xr-x.  1 root root   84 Sep 28 18:49 logs
drwxr-xr-x.  1 root root   84 Sep 28 18:49 temp
drwxr-xr-x.  9 root root 4096 Sep 28 18:49 web
root@localhost:/opt/sonarqube#
```

Please disregard this PR.

PR #115, opened 2017-10-21
Elasticsearch in SonarQube 6.6 requires non-root user and bash

Godin:


jan-z:
Shouldn't the volume get chown'ed to the sonarqube user in the run script? To enable upgrade from older versions.

Godin:
@jan-z `chown` added into run scripts

jan-z:
@Godin thanks, works fine for me now. Upgrade is working as well. LGTM

Godin:
@jan-z thank for reviewing and testing! :thumbsup:

PR #128, opened 2017-12-04
Removed blank lines

randyr505:
Blank lines where previous line has line continuation will fail in future versions.

### DIFF
--- a/5.6.7/Dockerfile
+++ b/5.6.7/Dockerfile
@@ -25,13 +25,11 @@ RUN set -x \
     && gosu nobody true
 
 RUN set -x \
-
     # pub   2048R/D26468DE 2015-05-25
     #       Key fingerprint = F118 2E81 C792 9289 21DB  CAB4 CFCA 4A29 D264 68DE
     # uid                  sonarsource_deployer (Sonarsource Deployer) <infra@sonarsource.com>
     # sub   2048R/06855C1D 2015-05-25
     && gpg --keyserver ha.pool.sks-keyservers.net --recv-keys F1182E81C792928921DBCAB4CFCA4A29D26468DE \
-
     && cd /opt \
     && curl -o sonarqube.zip -fSL https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-$SONAR_VERSION.zip \
     && curl -o sonarqube.zip.asc -fSL https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-$SONAR_VERSION.zip.asc \


randyr505:
I don't believe there is an official announcement. However, the line continuation merges the following line with the current line. In your Dockerfile's there are a few that have empty lines following the line continuation. That is the same as not having a line continuation. It works the same in bash. 

I was only building 6.7 so I didn't notice the others. You should update all Dockerfiles where you have a line continuation (\) followed by a blank/empty line.

Warning I received when building:
docker-sonarqube$ ./build.sh
Sending build context to Docker daemon  4.608kB
[WARNING]: Empty continuation line found in:
    RUN set -x     && gpg --keyserver ha.pool.sks-keyservers.net --recv-keys F1182E81C792928921DBCAB4CFCA4A29D26468DE     && cd /opt     && curl -o sonarqube.zip -fSL https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-$SONAR_VERSION.zip     && curl -o sonarqube.zip.asc -fSL https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-$SONAR_VERSION.zip.asc     && gpg --batch --verify sonarqube.zip.asc sonarqube.zip     && unzip sonarqube.zip     && mv sonarqube-$SONAR_VERSION sonarqube     && chown -R sonarqube:sonarqube sonarqube     && rm sonarqube.zip*     && rm -rf $SONARQUBE_HOME/bin/*
[WARNING]: Empty continuation lines will become errors in a future release.
Step 1/10 : FROM openjdk:8

Godin:
@randyr505 I'm asking reference on official documentation exactly because behavior doesn't look completely consistent with bash since `GNU bash, version 4.4.12(1)-release (x86_64-apple-darwin15.6.0)` doesn't even allow following
```
#!/bin/bash

echo line 1 \
  # comment
  line 2
```
producing
```
line 1
./test.sh: line 5: line: command not found
```

If both forbidden in docker, then better to fix both and not just remove blank lines. If only blank lines forbidden, then might be better to replace them by comments. Etc.

randyr505:
I mentioned bash mainly for how line continuation works. Comments I believe are allowed based on docker alone. Here is a bug that was reported/fixed when it previously complained about comments. It mentions that it will still complain about empty lines.

https://github.com/docker/docker-ce/commit/f6d296e26e23bd3d38b6d1e8bab8b69036055d8a

bartvanandel:
FWIW, if empty space (i.e., a line with no command or comment) is desired for easier eyeballing, you can use a line with just a continuation character and nothing else. If you want a continuation character after the last statement as well (for consistency, who knows), you can add a final line with just a semicolon to finish the statement.

As an example, we're using something like this (commands anonymized) without any problems or warnings:

```Dockerfile
RUN echo "Configuring ..." \
    && mv /tmp/foo/* /opt/foo \
    && mv /tmp/bar/* /opt/bar/ \
    \
    && ln -s /opt/foo/baz /opt/bar/qux \
    \
    # Apply patches (they must use absolute paths for this to work)
    && for patch in $(find /tmp/patches/ -name '*.patch'); do \
        patch -d/ -p0 -i "${patch}" \
        && chown boss:monster $(grep -o -E '^\+\+\+ \S*' "${patch}" | cut -b 5-) \
        || 2>/dev/null break 0; \
    done \
    \
    && rm -rf \
      /tmp/foo/ \
      /tmp/bar/ \
    \
    echo "Done." \
    ;
```

janos-ss:
Thank you for this contribution. I have added `7.4-community`,`7.4-community-alpine`, `6.7.6-community` and `6.7.6-community-alpine` in `master`, which also handle the changes suggested in this PR. Indeed, `docker build` commands were showing a warning about the blank lines, and that in later versions these will become errors.

PR #147, opened 2018-02-10
Use `restart: always` in docker-compose.yml

VonUniGE:
This docker-compose.yml works very well. The only problem is that when you have to restart the host (e.g. because of kernel upgrade) you have to run `docker-compose up` manually after the boot. Using `restart: always` allows the sonarqube service to survive host reboot without problems. Tested on Docker CE for Debian (Stretch).


pierre-guillot-sonarsource:
this behavior may not be wanted by everyone. The recipe docker-compose.yml is here to get users starting using docker-compose, not as a prod-ready bulletproof standard.

Feel free to fork the repo and customize the file to fulfil your needs.

PR #161, opened 2018-04-09
Update run.sh

chezuma:
The chown creates a permission denied when running in Kubernetes, since the chown is already done in the dockerfile it is not necessary in this script.


chezuma:
Seems to be an issue with GPG that fails randomly. :(

janos-ss:
Thanks. That was a good tip. This PR is outdated now, in current versions of the images we no longer do `chown` in the `run.sh` scripts.

PR #175, opened 2018-05-11
updated .yml for modern docker or docker UCP

ned-horvath:
Update to version 3 and use the default overlay network instead of an explicit one. 
I've used the revised version as a stack description with Docker EE as well as locally.


pierre-guillot-sonarsource:
No more traction on this PR for a while, i'm going to close it. If you feel this is still relevant, please describe why using an explicit network is wrong.

PR #183, opened 2018-06-23
Updated to 7.2

jincod:


janos-ss:
Thank you for this contribution. I have added `7.4-community`,`7.4-community-alpine`, `6.7.6-community` and `6.7.6-community-alpine` in `master`, with some other improvements, so I'm closing this PR now. I saw the issue you mentioned, and I solved it by passing `--no-tty` instead of `--batch`. Let me know if you think `--batch` would be better, and if so why.

PR #200, opened 2018-09-14
Add 7.3 docker image

CameronGo:
I've added the scripts needed for the 7.3 support, copied from previous versions.  I've tested the build and run processes with the exception of the gpg retrieval of the public key and the gpg verify, which weren't working for me due to the key server being unavailable.

Support for 7.2+ is important for us as part of our pilot deployment due to our requirement for Go support and I'd like to avoid building a custom image if possible.

If I can do anything else to help get a 7.3 image available, please let me know.

### DIFF
--- a/7.3-alpine/Dockerfile
+++ b/7.3-alpine/Dockerfile
@@ -1,0 +1,43 @@
+FROM openjdk:8-alpine
+
+ENV SONAR_VERSION=7.3 \
+    SONARQUBE_HOME=/opt/sonarqube \
+    # Database configuration
+    # Defaults to using H2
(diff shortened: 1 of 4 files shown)


ph-One:
While waiting for a 7.2 docker image, 7.3 dropped. Please make these images part of your release process, as we do not want to build our own image and deviate from what you all are doing, but we will if we're forced to.

janos-ss:
Thank you for this contribution. I have added `7.4-community`,`7.4-community-alpine`, `6.7.6-community` and `6.7.6-community-alpine` in `master`, with some other improvements, so I'm closing this PR now.

PR #205, opened 2018-09-27
Fallback to IPv4 pool to fix keyserver receive failed

tomverin:
Fix the random failure while receiving GPG key. Error comes from SKS Keyservers high-availability pool containing some IPv6-only servers

nodejs/docker-node#380 

Fix by adding a conditional expression to fallback to the IPv4-only SKS pool if we get an IPv6 server from the high-availability pool.


Godin:
@tomverin even if was merged, not sure that this is enough:

```
+ gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4
gpg: keybox '/tmp/tmp.35MM0H72yM/pubring.kbx' created
gpg: keyserver receive failed: Cannot assign requested address
+ gpg --keyserver ipv4.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4
gpg: keyserver receive failed: No data
```

PR #209, opened 2018-11-19
add missing --batch option to gpg command

KengoTODA:
When we build container image on Travis CI, [build failed because we have no `'/dev/tty'` in CI environment](https://travis-ci.org/spotbugs/sonar-findbugs/jobs/456801999). I also found that [we can avoid this problem by adding `--batch` option to `gpg` command](https://travis-ci.org/spotbugs/sonar-findbugs/jobs/456825859).

So to support the build in tty-less environment, I suggest to add `--batch` option to all `gpg` commands in Dockerfiles.

### DIFF
--- a/5.6.7-alpine/Dockerfile
+++ b/5.6.7-alpine/Dockerfile
@@ -23,8 +23,8 @@ RUN set -x \
     #       Key fingerprint = F118 2E81 C792 9289 21DB  CAB4 CFCA 4A29 D264 68DE
     # uid                  sonarsource_deployer (Sonarsource Deployer) <infra@sonarsource.com>
     # sub   2048R/06855C1D 2015-05-25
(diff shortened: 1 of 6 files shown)


KengoTODA:
CI build failed, but this is not related with the change in this PR.

> gpg: keyserver receive failed: Host is unreachable

janos-ss:
Thank you for this contribution. I have added `7.4-community`,`7.4-community-alpine`, `6.7.6-community` and `6.7.6-community-alpine` in `master`, with some other improvements, so I'm closing this PR now. I saw the issue you mentioned, and I solved it by passing `--no-tty` instead of `--batch`. Let me know if you think `--batch` would be better, and if so why.

KengoTODA:
Thanks for your confirmation!  
I think `--no-tty` is OK for this case. I choose `--batch` to keep consistency: it's already used in existing lines in Dockerfile.

PR #212, opened 2018-11-20
Feat: push to dockerhub registry

thecampagnards:
To automatically push on the registry <https://hub.docker.com/_/sonarqube/> the new images.

Need to add this env in travis :
- DOCKER_USERNAME
- DOCKER_PASSWORD

Which has access to the repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ script:
   - cd "$v"
   - docker build -t "sonarqube-$v" .
   - docker images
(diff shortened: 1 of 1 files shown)


janos-ss:
Thank you for this contribution.

I haven't seen anywhere in the Docker Hub documentation about pushing images there directly. I'm under the impression that the maintainers review requests for image updates in PRs they receive on [`docker-library/docs`](https://github.com/docker-library/docs) and [`docker-library/official-images`](https://github.com/docker-library/official-images), and they build and push images themselves. Please correct me if I'm wrong.

As for pushing images directly by CI, I don't think we want to do that. We prefer releases to be triggered by deliberate, explicit actions.

PR #215, opened 2018-11-22
MMF-1449 Provide a docker image for SonarQube community edition

janos-ss:


julienlancelot:
It's probably already somewhere, SONARQUBE_JDBC_USERNAME env variable will be deprecated ?

PR #225, opened 2018-12-06
Fix build after drop of Alpine-based images

derkoe:
Fix Travic-CI build after dropping the Alpine-based images in commit [82bbf8c756cfaaab6126d9879239729c0e833194]


Godin:
Already fixed in b5764382c51063841d24dbe33e18cf605fc1f08d

PR #232, opened 2019-01-02
Add JVM parameters on the wrapper

ulrichgi:
While trying to set memory parameters on both container and jvms for our sonar instances we realized that the wrapper can't be limited via Xmx parameter, and thus, in some cases was using more than 700Mo rss in the container.

To mitigate this issue, the workaround was to set JAVA_TOOL_OPTIONS environment variable.

This pull request adds the WRAPPER_JVM_ARGS environment variable in the run.sh script, which adds the capability to control the wrapper jvm size

### DIFF
--- a/6.7.6-community/run.sh
+++ b/6.7.6-community/run.sh
@@ -21,7 +21,7 @@ do
     fi
 done < <(env)
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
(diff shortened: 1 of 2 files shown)


simonbrandhof:
The memory of the Java Service Wrapper does not need to be increased. Consumption is supposed to be constant. 
The application memory should be tuned with the parameters `sonar.web.javaOpts` and `sonar.ce.javaOpts` (see [SonarQube documentation](https://docs.sonarqube.org/latest/setup/install-server/)).

Issue #240, opened 2019-01-31
issue while sonarqube running in conatiner with volume 


deekshith-elear:
Hi,
when i was trying to run sonarqube latest community edition in docker container with volume mounted to it. container is exited with below logs

2019.01.31 07:35:38 INFO  app[][o.s.a.AppFileSystem] Cleaning or creating temp directory /opt/sonarqube/temp
2019.01.31 07:35:38 INFO  app[][o.s.a.es.EsSettings] Elasticsearch listening on /127.0.0.1:9001
2019.01.31 07:35:38 INFO  app[][o.s.a.p.ProcessLauncherImpl] Launch process[[key='es', ipcIndex=1, logFilenamePrefix=es]] from [/opt/sonarqube/elasticsearch]: /opt/sonarqube/elasticsearch/bin/elasticsearch -Epath.conf=/opt/sonarqube/temp/conf/es
2019.01.31 07:35:39 INFO  app[][o.s.a.SchedulerImpl] Waiting for Elasticsearch to be up and running
2019.01.31 07:35:39 INFO  app[][o.e.p.PluginsService] no modules loaded
2019.01.31 07:35:39 INFO  app[][o.e.p.PluginsService] loaded plugin [org.elasticsearch.transport.Netty4Plugin]
2019.01.31 07:35:42 WARN  app[][o.s.a.p.AbstractProcessMonitor] Process exited with exit value [es]: 1
2019.01.31 07:35:42 INFO  app[][o.s.a.SchedulerImpl] Process [es] is stopped
2019.01.31 07:35:42 INFO  app[][o.s.a.SchedulerImpl] SonarQube is stopped

can anyone help me out to solve this issue

marksamodien:
I'm having the exact same issue.

deekshith-elear:
try by changing volume directory permissions to 766

marksamodien:
I just tried and still getting the error:

2019.02.01 12:53:30 INFO  app[][o.s.a.SchedulerImpl] Waiting for Elasticsearch to be up and running
2019.02.01 12:53:30 INFO  app[][o.e.p.PluginsService] no modules loaded
2019.02.01 12:53:30 INFO  app[][o.e.p.PluginsService] loaded plugin [org.elasticsearch.transport.Netty4Plugin]
2019.02.01 12:53:31 WARN  app[][o.s.a.p.AbstractProcessMonitor] Process exited with exit value [es]: 1
[...]

PR #246, opened 2019-02-16
Add custom PUID & PGID 

Yopii:
Hi,

I suggest this evolution which allows you to manage UID and GID via environment variables and to correct the problem of mounted volumes by the same occasion (#240).

Usage: `docker run --name sonar --detach --publish 9000:9000 --volume <mount-data>:/opt/sonarqube/data --env PUID=996 --env PGID=995 sonarqube`

No problem in volume mounted or data volume according to my tests. The process also runs with the user `sonarqube`.

If some people can test on their own (Kubernetes, etc.).

Thank you.

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] If the PR modifies or adds images, please use the `run-tests.sh imagedir` command to verify the image works
- [x] If the PR modifies or adds images, please try to make sure the images run correctly on Linux

### DIFF
--- a/6.7.6-community/run.sh
+++ b/6.7.6-community/run.sh
@@ -14,6 +14,12 @@ fi
 
 declare -a sq_opts
 
+# Update UID:GID for sonarqube user.
+usermod -u $PUID sonarqube && \
+groupmod -g $PGID sonarqube && \
+chown -R -h $PUID:$PGID $SONARQUBE_HOME && \
+usermod -g $PUID sonarqube
+
 while IFS='=' read -r envvar_key envvar_value
 do
     if [[ "$envvar_key" =~ sonar.* ]]; then
@@ -21,7 +27,8 @@ do
     fi
 done < <(env)
 
-exec java -jar lib/sonar-application-$SONAR_VERSION.jar \
+# Add call to gosu to drop from root user to sonarqube user.
+gosu sonarqube java -jar lib/sonar-application-$SONAR_VERSION.jar \
   -Dsonar.log.console=true \
   -Dsonar.jdbc.username="$SONARQUBE_JDBC_USERNAME" \
   -Dsonar.jdbc.password="$SONARQUBE_JDBC_PASSWORD" \
--- a/6.7.6-community/Dockerfile
+++ b/6.7.6-community/Dockerfile
@@ -1,6 +1,8 @@
 FROM openjdk:8
 
 ENV SONAR_VERSION=6.7.6 \
+    PUID=999 \
+    PGID=999 \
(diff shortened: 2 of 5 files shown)


pierre-guillot-sonarsource:
Hi, thanks for your PR. Just to be sure to understand the problem : the permission troubles appear if you created the volume with the SQ alpine-based image, and then switch to the SQ debian-based image ?

pjmpsu:
No.  The issue is that the polkitd user on Centos has the same UID/GID as the sonarqube user in the container and causes issues when mounting a volume for data.

pierre-guillot-sonarsource:
You mean your docker host is Centos ?

pjmpsu:
That is correct.  Our docker hosts are Centos based.

isuftin:
@Yopii it seems like there are conflicting files for this PR. I'm not sure if this PR is still being looked at but could you clean up the conflicts? I hope this to be pulled in soon as I am having the same issue and I will have to write my own Sonar Docker image which incorporates these changes if this isn't pulled and released soon :(

6.7.6-community/run.sh
```diff
@@ -14,14 +14,21 @@ fi
 
 declare -a sq_opts
 
+# Update UID:GID for sonarqube user.
+usermod -u $PUID sonarqube && \
+groupmod -g $PGID sonarqube && \
+chown -R -h $PUID:$PGID $SONARQUBE_HOME && \
+usermod -g $PUID sonarqube
```
skyghis: `usermod -g` changes the group; Should be `$PGID` ?
Same suggestion goes for `7.6-community/run.sh` at line 21

```suggestion
usermod -g $PGID sonarqube
```

6.7.6-community/run.sh
```diff
@@ -14,14 +14,21 @@ fi
 
 declare -a sq_opts
 
+# Update UID:GID for sonarqube user.
+usermod -u $PUID sonarqube && \
+groupmod -g $PGID sonarqube && \
```
isuftin: Gets permission denied for this

6.7.6-community/run.sh
```diff
@@ -14,14 +14,21 @@ fi
 
 declare -a sq_opts
 
+# Update UID:GID for sonarqube user.
+usermod -u $PUID sonarqube && \
```
isuftin: If using an uid that the image was not built with, we get

```
usermod: user sonarqube is currently used by process 1
```

here

pierre-guillot-sonarsource:
I'm closing this as it is duplicate of #344 which is more recent & active

PR #259, opened 2019-03-21
Add env support for ldap properties

jkroepke:
Since the ldap properties are not starting with sonar. the env vars will be ignored.

https://docs.sonarqube.org/latest/instance-administration/delegated-auth/#header-2


simonbrandhof:
It makes sense. Thanks for the contribution.

PR #272, opened 2019-06-11
Dockerfile for 7.7 Developer Edition

DuAell:
Dockerfile for SonarQube 7.7 Developer edition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ services:
 env:
   - VERSION=6.7.7-community
   - VERSION=7.7-community
+  - VERSION=7.7-developer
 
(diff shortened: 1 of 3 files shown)


jcoeltjen:
The `run.sh` is not executable when I clone your repository.
This prevents the created container from starting up on my machine.

DuAell:
Saw that.
This is the exact same one than in folder 7.7-community

I don't know why this is not working, and unfortunately, don't have time to investigate.

jcoeltjen:
As I cannot add commits to your PR I created another PR with your changes and set the executable bit for `run.sh`. #273

PR #273, opened 2019-06-12
Dockerfile for 7.7 Developer Edition (continuation)

jcoeltjen:
Continuation of #272.
Only change I added is to set the executable bit for `run.sh`.

Credit for the dockerfile itself belongs to the original author.


jcoeltjen:
Closing in favor of #280

PR #293, opened 2019-07-23
MMF-1777 add healcheck on api/system/status

pierre-guillot-sonarsource:


pierre-guillot-sonarsource:
Checklist :
- test db migration
- verify that the API call is doing a SQL request

pierre-guillot-sonarsource:
As it is not recommended to have healthcheck on official image, we provided an example of how it can be implemented here : https://github.com/docker-library/healthcheck/pull/20

PR #297, opened 2019-07-24
MMF-1777 display elasticsearch logs

michal-duda-sonarsource:


michal-duda-sonarsource:
Should be better now.

pierre-guillot-sonarsource:
tested, LGTM

PR #299, opened 2019-07-26
MMF-1777 clean up dockerfile

michal-duda-sonarsource:


7.9.1-community/Dockerfile
```diff
@@ -8,31 +8,13 @@ ENV SONAR_VERSION=7.9.1 \
...
-                            pgp.mit.edu) ; do \
-        gpg --batch --keyserver "$server" --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4 && break || : ; \
-    done \
-    && gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu \
-    && rm -rf "$GNUPGHOME" /usr/local/bin/gosu.asc \
-    && chmod +x /usr/local/bin/gosu \
-    && gosu nobody true
-
 RUN set -x \
+    && groupadd -r sonarqube && useradd -r -g sonarqube sonarqube \
```
pierre-guillot-sonarsource: for better layer cacheability, i would leave this in a separate layer : there is no need to redo it if the SQ binary change

7.9.1-community/Dockerfile
```diff
@@ -8,31 +8,13 @@ ENV SONAR_VERSION=7.9.1 \
...
     # pub   2048R/D26468DE 2015-05-25
     #       Key fingerprint = F118 2E81 C792 9289 21DB  CAB4 CFCA 4A29 D264 68DE
     # uid                  sonarsource_deployer (Sonarsource Deployer) <infra@sonarsource.com>
```
pierre-guillot-sonarsource: we may even want to have the gpg thing on a separate layer as well, cutting this in two, after line 28 (`done \`)

7.9.1-community/Dockerfile
michal-duda-sonarsource: 👍

7.9.1-community/Dockerfile
michal-duda-sonarsource: 👍

PR #301, opened 2019-07-29
add daily tests for public image

pierre-guillot-sonarsource:


pierre-guillot-sonarsource:
> LGTM but I'm no bash expert :-)

the script is a copy/paste of run-tests.sh, with the source of the image hardcoded to `sonarqube`

PR #302, opened 2019-07-29
MMF-1777 switch to slim version of openjdk image

michal-duda-sonarsource:


7/community/Dockerfile
```diff
@@ -1,20 +1,11 @@
...
-
-# Http port
-EXPOSE 9000
-
-RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
+ARG SQ_VERSION=7.9.1
 
+FROM openjdk:11-slim AS build
+ARG SQ_VERSION
+ENV SONAR_VERSION=${SQ_VERSION} \
```
pierre-guillot-sonarsource: do we really need the SQ_VERSION in the build step ? if we can hardcode a folder, it is simpler

michal-duda-sonarsource: In my opinion having a version defined in only one place in `Dockerfile` is the best way to avoid typos or version mismatches in multi-stage build. I would be for hardcoding if version was only used in one place.

PR #303, opened 2019-07-31
MMF-1777 add comment about the tail workaround on the es.log

pierre-guillot-sonarsource:


7/community/run.sh
```diff
@@ -21,7 +21,7 @@ do
     fi
 done < <(env)
 
-exec tail -F ./logs/es.log &
+exec tail -F ./logs/es.log & # this tail on the elasticsearch logs is a temporary workaround, see https://github.com/docker-library/official-images/pull/6361#issuecomment-516184762
```
michal-duda-sonarsource: I think he meant reference to our ticket

pierre-guillot-sonarsource: i mean, but i found his response "better" in a way that it explain why this is not a best practice.

Issue #323, opened 2019-09-26
JDBC settings from sonar.properties are ignored


sns-seb:
# Observed behavior

When mounting the `conf` directory with a customized `sonar.properties` file which defines the JDBC properties to connect to the database, SonarQube ignores these properties.

# Expected behavior

SonarQube connects to the database I specify by properties in the `sonar.properties` file.

# How to reproduce

(with SQ 8 community built from branch [feature/sl/initialized_read_volumes](https://github.com/SonarSource/docker-sonarqube/tree/feature/sl/initialized_read_volumes))

Create a `conf` directory to mount in SQ container:

```
mkdir -p /tmp/sq/conf
chmod 777 /tmp/sq/conf
```

Start SQ container to get an initialized conf directory. Observe that SQ is running on H2.
```
docker run -p "9000:9000" --name="sq" -v "/tmp/sq/conf:/opt/sonarqube/conf" -d sq8
```

Stop SQ:

```
docker stop sq
```

Add JDBC properties to `sonar.properties`:

```
sudo vi /tmp/sq/conf/sonar.properties
```

Sample jdbc properties, exact values do not matter since they are ignored:

```
sonar.jdbc.username=foo
sonar.jdbc.password=bar
sonar.jdbc.url=jdbc:postgresql://localhost:5432/acme
```

Restart SQ container:

```
docker start sq
```

Connect to SQ and/or check logs, SQ still connect to an H2 database.

# Suspected cause

Entry point script `run.sh` passes env properties to the SonarQube process for JDBC properties `sonar.jdbc.name`, `sonar.jdbc.password` and `sonar.jdbc.url`.

These properties takes precedence over the content of `sonar.properties` *even when empty*.

In the scenario above, `sonar.jdbc.url` is empty because its value is the one of the env variable `SONAR_JDBC_URL`. Even when `SONAR_JDBC_URL` is not set, this defines an empty value to the env property.

An empty `sonar.jdbc.url` is interpreted by SonarQube as "connect to H2".

Note that as a consequence, the same problem applies to property `sonar.web.javaAdditionalOpts`.

QkiZMR:
This bug is still present in 7.9.4-community version

PR #327, opened 2019-09-30
fix sonar.properties properties ignored

sns-seb:
This PR fixes issue #323 

- [x] fix `sonar.jdbc` properties
- [x] fix `sonar.web.javaAdditionalOpts`
    - this property was defined to make web process use /dev/urandom instead of /dev/random by setting env property `java.security.egd`
   - there is no simple way set this property via `sonar.web.javaAdditionalOpts` and at the same time no overwrite `sonar.web.javaAdditionalOpts` defined in `sonar.properties`
   - this has been worked around by setting `securerandom.source` at Java installation level (which affects all Java processes in the container) and not at all anymore at the SQ Web process level only

### DIFF
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -18,6 +18,8 @@ EXPOSE 9000
 RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
 
 SHELL ["/bin/bash", "-c"]
+RUN sed -i -e "s/securerandom.source=file:\/dev\/random/securerandom.source=file:\/dev\/urandom/g" \
+  $JAVA_HOME/conf/security/java.security
 RUN set -x \
     && cd /opt \
     && if [ "x$SONARQUBE_ZIP_USERNAME" = "x" ] ; \
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -6,10 +6,7 @@ ARG SONARQUBE_ZIP_URL=https://${SONARQUBE_ZIP_SERVER}/Distribution/sonarqube/son
 ARG SONARQUBE_ZIP_USERNAME
 ARG SONARQUBE_ZIP_PASSWORD
 ENV SONAR_VERSION=${SONARQUBE_VERSION} \
-    SONARQUBE_HOME=/opt/sonarqube \
-    SONARQUBE_JDBC_USERNAME=sonar \
-    SONARQUBE_JDBC_PASSWORD=sonar \
-    SONARQUBE_JDBC_URL=""
+    SONARQUBE_HOME=/opt/sonarqube
 
 RUN apt-get update \
     && apt-get install -y curl unzip \
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -18,8 +18,8 @@ EXPOSE 9000
 RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
 
 SHELL ["/bin/bash", "-c"]
-RUN sed -i -e "s/securerandom.source=file:\/dev\/random/securerandom.source=file:\/dev\/urandom/g" \
-  $JAVA_HOME/conf/security/java.security
+RUN sed -i -e "s?securerandom.source=file:/dev/random?securerandom.source=file:/dev/urandom?g" \
+  "$JAVA_HOME/conf/security/java.security"
 RUN set -x \
     && cd /opt \
     && if [ "x$SONARQUBE_ZIP_USERNAME" = "x" ] ; \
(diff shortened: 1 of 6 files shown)


sns-seb:
FYI, @janos-ss , thanks to your suggestion, I've enabled bash strict mode in this PR

8/community/run.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-set -e
+set -eou pipefail
```
janos-ss: I'm a bit surprised this worked, since `pipefail` is a parameter of `-o`. I would write `set -euo pipefail` because it's less surprising.

8/developer/run.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-set -e
+set -eou pipefail
```
janos-ss: Same comment as in the other file.

8/enterprise/run.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-set -e
+set -eou pipefail
```
janos-ss: Same comment ;-)

8/community/Dockerfile
```diff
@@ -21,6 +18,8 @@ EXPOSE 9000
 RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
 
 SHELL ["/bin/bash", "-c"]
+RUN sed -i -e "s/securerandom.source=file:\/dev\/random/securerandom.source=file:\/dev\/urandom/g" \
+  $JAVA_HOME/conf/security/java.security
```
janos-ss:
```suggestion
  "$JAVA_HOME/conf/security/java.security"
```

8/community/Dockerfile
```diff
@@ -21,6 +18,8 @@ EXPOSE 9000
 RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
 
 SHELL ["/bin/bash", "-c"]
+RUN sed -i -e "s/securerandom.source=file:\/dev\/random/securerandom.source=file:\/dev\/urandom/g" \
```
janos-ss: Have you considered using an alternative pattern separator of `s///`, for example `s???` or `s|||` ?
```suggestion
RUN sed -i -e "s?securerandom.source=file:/dev/random?securerandom.source=file:/dev/urandom?g" \
```

8/community/run.sh
```diff
@@ -1,31 +1,38 @@
...
-if [ "${1:0:1}" != '-' ]; then
-  exec "$@"
+if [[ "${1:-}" != -* ]]; then
+  exec "$@" 
 fi
 
+declare -a sq_opts
+
+add_env_var_as_env_prop() {
+  if [ ! -z "$1" ]; then
```
janos-ss: I believe this is identical, shorter, and quite intuitive:
```suggestion
  if [ "$1" ]; then
```

8/community/run.sh
```diff
@@ -1,31 +1,38 @@
...
 do
     if [[ "$envvar_key" =~ sonar.* ]] || [[ "$envvar_key" =~ ldap.* ]]; then
         sq_opts+=("-D${envvar_key}=${envvar_value}")
     fi
 done < <(env)
+# map legacy env variables
+add_env_var_as_env_prop "${SONARQUBE_JDBC_USERNAME:-}" "sonar.jdbc.username"
+add_env_var_as_env_prop "${SONARQUBE_JDBC_PASSWORD:-}" "sonar.jdbc.password"
+add_env_var_as_env_prop "${SONARQUBE_JDBC_URL:-}" "sonar.jdbc.url"
+add_env_var_as_env_prop "${SONARQUBE_WEB_JVM_OPTS:-}" "sonar.web.javaAdditionalOpts"
```
janos-ss: Since we usually assign to the left side things on the right side, I find this usage here a bit confusing. I would replace `add_env_var_as_env_prop` with `set_prop_from_env_var` so that I could write:
```
set_prop_from_env_var "sonar.web.javaAdditionalOpts" "${SONARQUBE_WEB_JVM_OPTS:-}"
```

8/developer/run.sh
```diff
@@ -1,31 +1,38 @@
...
 
-if [ "${1:0:1}" != '-' ]; then
-  exec "$@"
+if [[ "${1:-}" != -* ]]; then
+  exec "$@"
```
janos-ss: Trailing garbage space?

8/community/run.sh
```diff
@@ -1,31 +1,38 @@
 #!/usr/bin/env bash
 
-set -e
+set -eou pipefail
 
-if [ "${1:0:1}" != '-' ]; then
-  exec "$@"
+if [[ "${1:-}" != -* ]]; then
+  exec "$@"
```
janos-ss: Trailing garbage space?

8/enterprise/run.sh
```diff
@@ -1,31 +1,38 @@
...
 
-if [ "${1:0:1}" != '-' ]; then
-  exec "$@"
+if [[ "${1:-}" != -* ]]; then
+  exec "$@"
```
janos-ss: Trailing garbage space?

8/community/Dockerfile
sns-seb: I didn't know this one even a feature. I hardly did more than make my own sed from various sources on the web and all use `|`

8/community/run.sh
pierre-guillot-sonarsource: done 👍

PR #330, opened 2019-10-07
don't use env vars to pass sensitive information

sns-seb:


8/docker-build.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
michal-duda-sonarsource: Think we should rename this file or make it clear that it's only to be used for building using private repos. Or make private server env vars optional for this script.

sns-seb: I take it you don't find the documentation inside the script to be enough?

Do you have a name to suggest? I don't have anything satisfying coming to mind

michal-duda-sonarsource: how about `build-docker-from-private-repo.sh`?

PR #343, opened 2019-11-04
Update developer documentation with new 8.0 images

sns-seb:
- [X] explain directory structure by edition
- [X] explain new (and secured) way to build images from non public artifacts
- [X] explain branch "branch-8"
- [X] add details about ITs
- [X] update release guidelines

### DIFF
--- a/develop.md
+++ b/develop.md
@@ -1,49 +1,195 @@
-Develop
-=======
+Guidelines and documentation for developers of the SonarQube Docker images.
 
-Helpful notes for developers of the images.
 
-Releasing a new version of an image
------------------------------------
+Adding images for a new version of SQ
+=====================================
 
-Create the new image for the new version:
+New major version
+-----------------
 
-1. Rename the last version's directory, for example `7.8-community` to `7.9-community`
-2. Update the version in `Dockerfile`, and make any other necessary changes (usually none)
-3. Update the version in `.travis.yml`
-4. Run tests, for example: `./run-tests.sh 7.9-community`
-5. Create a Pull Request, confirm Travis build passes, get it reviewed and merged
+Create a new subdirectory with the major version number in the root directory of the repository with a sub directory for each supported edition, eg.: 
 
-Update the public repositories for [Docker Hub][hub]:
+```
+mkdir -p 9/community 9/developer 9/enterprise
+```
 
-1. Only if necessary, update files in https://github.com/docker-library/docs/tree/master/sonarqube and create a Pull Request
-    1. The markdown format must follow a certain standard, otherwise automated tests will fail. You can test with the `markdownfmt.sh` tool included in the repository, for example `./markdownfmt.sh -d sonarqube/content.md` will output the diff that would have to be done to make the tests pass. You can use the `patch` command to apply the changes, for example: `./markdownfmt.sh -d sonarqube/content.md | patch sonarqube/content.md`
-    2. Verify the Pull Request passes the automated tests (visible in the status of the PR)
+Each edition directory will contain a single Dockerfile.
 
-2. Update https://github.com/docker-library/official-images/blob/master/library/sonarqube
-    1. Update the `Tags` and `Directory` entries appropriately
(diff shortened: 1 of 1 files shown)


sns-seb:
@pierre-guillot-sonarsource since you created the doc PR, do you see any other relevant information/recommendation to add to the part about it?

develop.md
```diff
@@ -1,49 +1,195 @@
...
-    4. Verify the Pull Request passes the automated tests (visible in the status of the PR)
+New non-major version
+---------------------
 
-3. Wait for the Pull Requests to get merged, and verify the updated page on [Docker Hub][hub]
+As of today, we publish only a single version of SQ for a given major version in the "master" branch.
 
-FAQ
----
+New new image is therefor created for non-major versions of SQ. Instead, existing images are updated.
```
pierre-guillot-sonarsource:
```suggestion
New image is therefor created for non-major versions of SQ. Instead, existing images are updated.
```

michal-duda-sonarsource:
```suggestion
New new image is therefore created for non-major versions of SQ. Instead, existing images are updated.
```

michal-duda-sonarsource: `Instead`?

develop.md
```diff
@@ -1,49 +1,195 @@
...
+=========================================
+
+Dogfooding
+----------
+
+SonarQube Docker images are dogfooded.
+
+This implies that Docker images are produced for every artifact produced from the dogfood branch of SonarQube (see the [dogfood_docker_build_task](https://github.com/SonarSource/sonar-enterprise/blob/master/.cirrus.yml#L263)).
+
+Docker images of SonarQube are built from publicly available artifacts, which means the DockerFile can simply download them from the public place they are hosted (`binaries.sonarsource.com`).
```
michal-duda-sonarsource: It's always named just`Dockerfile`, no need to uppercase the `file` part (needs changing in a few places).

develop.md
```diff
@@ -1,49 +1,195 @@
...
------------
+Dogfooding and images for local artifacts
+=========================================
+
+Dogfooding
+----------
+
+SonarQube Docker images are dogfooded.
+
+This implies that Docker images are produced for every artifact produced from the dogfood branch of SonarQube (see the [dogfood_docker_build_task](https://github.com/SonarSource/sonar-enterprise/blob/master/.cirrus.yml#L263)).
```
michal-duda-sonarsource: Link is quite useful now but will become outdated very quickly. Not a big deal and I have no strong opinion but maybe just name of the job + the link to cirrus `conf` would be better?

develop.md
```diff
@@ -1,49 +1,195 @@
...
+4. an IT ensuring that the 6 DockerFiles are up to date with the "template" and the script (to prevent dev from forgetting to commit up-to-date DockerFiles)
+
+With this idea:
 
-- https://github.com/docker-library/docs
-- https://github.com/docker-library/official-images
+1. trust in dogfooding Docker images representatives of the one which will end up as the official images moves from the developer doing the right changes when cleaning the "version" branch to the script and the template
+  * no more human based last minute changes is an obvious improvement
+  * it's all commited so an error is easy to track
+2. there is no more need for a "version" branch
```
michal-duda-sonarsource: `... no longer a need ...`

develop.md
sns-seb: "New new" was intended to be "No new"

Indeed, with that typo, both sentences hardly make any sense

sns-seb: "New new" should actually be "No new"

fixing

develop.md
sns-seb: I thought about it

I considered that the link is correct now and this is convenient.
If line becomes incorrect in the future, the link will still point to the correct file. A bit less convenient, the result will be the same as providing the link without the line but at least it's been more convenient for a while

sns-seb: I thought about it

I considered that the link is correct now and this is convenient.
If line becomes incorrect in the future, the link will still point to the correct file. A bit less convenient, the result will be the same as providing the link without the line but at least it's been more convenient for a while

PR #344, opened 2019-11-04
Update 7 to be in line with 8 builds, allow custom UID/GID

isuftin:
This PR updates the building of the Docker image for version 7 to be in-line with the processing that happens with the 8/* images. This means that we now have an internal SonarQube dir that creates symlinks to a public dir and copies files there. 

I've added empty zip/ dirs to the 7 and 8/* subdirs so the COPY command does not fail when building. If a sonarqube zip file is in there, it works as expected. Otherwise, it's essentially a no-op besides copying in an empty ".keep" file. 

This update also allows the builder to specify a custom UID and GID as in https://github.com/SonarSource/docker-sonarqube/pull/246 - This allows the containers to run on systems that already have a uid or gid of 999. 

The run script for version 7 also gets updated to be in line with the run scripts for versions 8/*


pierre-guillot-sonarsource:
Hi, thanks for your contribution. I'm setting up a CentOS env to reproduce and will review your PR. Please consider submitting two different PR :
- One for the addition of custom UID/GID (this one)
- One for backporting all Dockerfile changes to the 7.9 LTS version

On the second point, i'm not sure though that this contribution will make it to docker hub. Our last stand about the 7.X docker version was to not support them, and to not update them. But we will discuss the idea internally.

isuftin:
@pierre-guillot-sonarsource One question I do have here. The 8* builds are labeled as 'beta' so I was under the assumption that the 7 build is the LTS build and in which case it's still actively developed. What is the suggestion from the SonarSource team on the build to use for our production instances?

pierre-guillot-sonarsource:
@isuftin the 8.X branch will be labelled as beta until 8.1 or 8.2, until we gathered enough feedback and adjust accordingly.

The SonarQube 7.9.X branch is LTS, receiving only security fix and blocker bugfix. It is actively `supported` until the next LTS, but not actively `developed`. As we want to heavily use internally each version we release, it's not under our radar to make big changes to the 7.9.X docker version.

About your production instances, i would say it depend on your organisation maturity with Docker and what edition you are using. If your all infrastructure is already running on Docker and you are comfortable with it, go for the 8.0 beta version. The `beta` tag stand for the Docker part : SonarQube 8.0 is an official release.

pierre-guillot-sonarsource:
I tested this on debian & centos, the issue i got is that the `chown` you added in the `run.sh` can only work if the container is run with `--user=root`. This may be an acceptable condition if well documented. And in this case i guess we want to do the chown only when the `--init` flag is set

isuftin:
@pierre-guillot-sonarsource Thanks for the testing and I will try to test the same on my end and provide some integration testing harness in this PR along with my findings. In the end, we will probably want to remove the chown in the run.sh completely and I will probably amend my PR to reflect that.

One thing I noticed is that even when not running with `--init`, the application does seem to initialize empty dirs if they're found at start. I may be mistalen and was doing something funky, but if that's the case, do we actually need `--init` at all?

isuftin:
Closing this PR and will re-open separate PRs if needed

pierre-guillot-sonarsource:
We are about to drop the `--init` command, and change our strategy on plugin initialisation at startup. You can have a look at https://jira.sonarsource.com/browse/MMF-1881, but this is still a draft. Maybe you want to wait for this subject to move on our side before spending time to adapt, if things are going to change soon.

PR #342, opened 2019-10-30
SONAR-12696 Add libfreetype6 and libfontconfig1 for svg-badges

sjaek:
The docker image currently does not allow the SVG badges plugin to be installed because that requires libfreetype and fontconfig.

Currently if you install the svg plugin in the docker image it will fail to start.

Added those to the Dockerfile and tested that the SVG plugin can be installed.

PR #347, opened 2019-11-11
Error Plugin SonarQube Badges

krahllucas:
This error ocurrs when install Badges on SonarQube, after restart. (Using Docker)
image


pierre-guillot-sonarsource:
Hello and thanks for submitting this PR. I think this is a duplicate of #342

Issue #223, opened 2018-12-01
Provide a docker image for SonarQube Developer Edition


mind-blowing:
Currently there are no docker image for `SonarQube Developer Edition`, it will be nice if you will support it in your official docker repository.

shadowXVll:
Also if anyone can provide some info about promotion from community to dev edition? In docker container. Thx in advance.

bendavidpaz:
I agree with the previous two comments.  An image with the developer edition pre-installed would be great. Also an explanation of how to map container data sets to docker volumes would be great.   I plan to use postgres for the DB - in a separate container.

Mitsufumi:
Yopii made great job. He prepared a docker image for SonarQube Developer Edition.
Thank you so much, Yopii!!

@bendavidpaz 
You might know it though, I don't recommend you to use DB in container. Container is ephemeral.

MarkWillems:
Where can I test this?

SeanTurner026:
Yopii did a great job. I built his docker image, modified the official helm chart values.yaml to use the image I built, navigated to /setup, upgraded my database, and everything works.

simonbrandhof:
For reference: https://jira.sonarsource.com/browse/MMF-1604

itshorty:
Is there already an expected release for the developer image?

claasd:
What is blocking this PR?

alexmngn:
Is there a way to upgrade the community edition docker image to use the developer edition?

jcoeltjen:
There is a PR for 7.7 Developer Edition: #273

ricardompcarvalho:
Does anyone know if there is an Enterprise Edition image?

bdavis-tc:
> Does anyone know if there is an Enterprise Edition image?

**Community Edition**: official docker image is readily available and actively maintained and updated
https://hub.docker.com/_/sonarqube

From what I've seen on this topic; the status for **Enterprise Edition** is the same as **Developer Edition**.

- There may be 3rd party dockerfiles you can use to build an image.
- You might even be able to use the Community Image and install the license plugin.

(I haven't attempted yet, but I will most likely be expected to with-in the next few months.)

Here are some links to discussions I've found that I consider to be related to this topic:
- https://community.sonarsource.com/t/sonarqube-docker-images/4209
- https://community.sonarsource.com/t/supported-docker-container-version-of-sonarqube-developer-enterprise-edition/8954
[...]

PR #352, opened 2019-11-21
Symlink the entrypoint script to the 7.9 location

SinisterMinister:
The stable SonarQube Helm chart directly refers to the entrypoint script defined in 7.9.1's image. Since this file moved in the new approach with 8.x, the images do not work for helm users. With the added symlink, this allows the helm charts to continue to function with the new images.

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] If the PR modifies or adds images, please use the `run-tests.sh imagedir` command to verify the image works
- [x] If the PR modifies or adds images, please try to make sure the images run correctly on Linux

### DIFF
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -53,6 +53,10 @@ RUN set -x \
 
 COPY --chown=sonarqube:sonarqube run.sh "$SONARQUBE_HOME/bin/"
 
+# Backwards compatibility for Helm Chart users
+RUN mkdir -p "$SONARQUBE_PUBLIC_HOME/bin" && \
+    ln -s "$SONARQUBE_HOME/bin/run.sh" "$SONARQUBE_PUBLIC_HOME/bin/run.sh" 
+
 USER sonarqube
 WORKDIR $SONARQUBE_HOME
 ENTRYPOINT ["./bin/run.sh"]
--- a/8/developer/Dockerfile
+++ b/8/developer/Dockerfile
@@ -53,6 +53,10 @@ RUN set -x \
 
 COPY --chown=sonarqube:sonarqube run.sh "$SONARQUBE_HOME/bin/"
 
+# Backwards compatibility for Helm Chart users
+RUN mkdir -p "$SONARQUBE_PUBLIC_HOME/bin" && \
+    ln -s "$SONARQUBE_HOME/bin/run.sh" "$SONARQUBE_PUBLIC_HOME/bin/run.sh" 
+
 USER sonarqube
 WORKDIR $SONARQUBE_HOME
 ENTRYPOINT ["./bin/run.sh"]
--- a/8/enterprise/Dockerfile
+++ b/8/enterprise/Dockerfile
@@ -53,6 +53,10 @@ RUN set -x \
 
 COPY --chown=sonarqube:sonarqube run.sh "$SONARQUBE_HOME/bin/"
 
+# Backwards compatibility for Helm Chart users
+RUN mkdir -p "$SONARQUBE_PUBLIC_HOME/bin" && \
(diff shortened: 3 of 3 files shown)


pierre-guillot-sonarsource:
Hell @sns-seb , what do you think about this ?

sns-seb:
@pierre-guillot-sonarsource I think it boils down to:

1. was the name and location of the entrypoint a public "API"
2. if not, do we want to make it a public "API"?
3. if so, is it the right name and location for that public "API"?

Also, I wonder whether, in general in the Docker work, endpoints are considered as public "API".

Finally, I'm concerned with adding patches to the 8.X images to provide compatibility with this and that usage of the 7.9.1 images.

I understand the situation with SonarQube Helm chart but I don't think we are bound to apply this fix (putting aside optimizing the Docker image):

1. 7.9.1-community image was provided with no warranty
2. SonarQube Helm chart takes the SQ docker image name as a parameter (and default value is 7.9.1-community), this issue is not a blocker. Helm users are just stuck to the 7.9.1-community if they want to use the stable SonarQube Helm chart.
3. a SonarQube Helm chart compatible with 8.0 or 8.1 onward could probably be released
4. "Stable" is a contract of Helm, not of SonarQube docker images

SinisterMinister:
> SonarQube Helm chart takes the SQ docker image name as a parameter (and default value is 7.9.1-community), this issue is not a blocker.

I have to take exception to this point. I wouldn't have submitted this PR if it was not a blocker for us. As it stands, there is no way for Docker or Kubernetes users to use a paid edition of SonarQube. There is mention of using the License Manager Plugin in the original ticket regarding paid edition images, but it seems like that's been removed from the Internet as I can't find a single download for it anywhere. If you guys have that information, it would be super helpful to communicate it in the docs for those like myself who are stuck in between incompatible releases.

> Finally, I'm concerned with adding patches to the 8.X images to provide compatibility with this and that usage of the 7.9.1 images.

I'm a little concerned that this is raising dogmatic issues regarding contract stability rather than pragmatic user experience stability. It took almost a year for #223 to be implemented and then it was done so in a way that was incompatible with the community tooling so that the users wishing to use it are unable to. I totally understand the need for the improvements but it would have been better for us users if it were done in a way that allowed adoption rather than split the user base into 7.9 users, 8.x users, and users like myself who are forced to extend the image and now are out of band completely to your process. I really hope this factors into the calculus as well.

> Also, I wonder whether, in general in the Docker world, entrypoints are considered as public "API".

I can't speak for the whole community but personally I consider the whole image to be a contract of sorts because people can and will extend your images and will expect things to be in certain places when use them. In this case, the helm chart needed to copy the plugins to the image before running SonarQube. Due to how Kubernetes works, this *overrides* the entrypoint script with the provided command. Since this happens, the end of that script must [make the call to the original entrypoint script](https://github.com/helm/charts/blob/master/stable/sonarqube/templates/copy-plugins.yaml#L28) itself. This is [standard Kubernetes behaviour](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes). Which leads me to my next quesion.

> in general in the Docker world

Docker is just one of many container run-time engines. Does this mean that you guys do not plan to support usage in Kubernetes, Mesos, DigitalOcean, AWS ECS and the like?

SinisterMinister:
After re-reviewing the helm chart, this could potentially be addressed in there with usage of an init container so as to not change the image command if you guys want to keep this PR out of the mix. That said, electing to go this route could potentially wind up with other PRs like this if there is other affected tooling out there.

Also, apologies if the tone in my previous message came across too strongly as it wasn't my intention. I have yet to have my morning coffee...

PR #360, opened 2019-12-23
SONAR-12696 Add libfreetype6 and libfontconfig1 for svg-badges backport to 7

BerndFarkaDyna:
with SONAR-12969 the fix was only backported to 8 but not 7..

... i really need to have to have the svg-badges on the current LTS version which is not possible ATM

### DIFF
--- a/7/community/Dockerfile
+++ b/7/community/Dockerfile
@@ -1,7 +1,7 @@
 FROM openjdk:11-jre-slim
 
 RUN apt-get update \
-    && apt-get install -y curl gnupg2 unzip \
+    && apt-get install -y curl gnupg2 unzip libfreetype6 libfontconfig1 \
     && rm -rf /var/lib/apt/lists/*
 
 ENV SONAR_VERSION=7.9.2 \


BerndFarkaDyna:
this is the port of pr #342 back to sonarqube 7 (LTS)

BerndFarkaDyna:
about the broken build: it doesn't look related to may change since there are problems in 8....

wouter-admiraal-sonarsource:
@pierre-guillot-sonarsource Internal PR for QA is #361

pierre-guillot-sonarsource:
> with SONAR-12969 the fix was only backported to 8 but not 7..
> ... i really need to have to have the svg-badges on the current LTS version which is not possible ATM

Hello, just out of curiosity, have you tried the built-in badge feature ?

BerndFarkaDyna:
@pierre-guillot-sonarsource: yes we did, but to keep backwards compatibility it would be great to have both...

pierre-guillot-sonarsource:
I understand that from a generic point of view, one can argue that keeping backward compatibility is nice. But in this specific case, the plugin has not been maintained for 3+ years, while SonarQube provide an equivalent built-in feature.

In the end, what make us add the library in the Dockerfile 8 is that it is also needed for the PDF feature of enterprise edition. Since the 7.9 LTS is community only for Docker, we don't have both need.

Is there a specific use case that you can not achieve with the built-in feature; that is accomplished by the svg-badge plugin ?

BerndFarkaDyna:
> I understand that from a generic point of view, one can argue that keeping backward compatibility is nice. But in this specific case, the plugin has not been maintained for 3+ years, while SonarQube provide an equivalent built-in feature.
> 
> In the end, what made us add the library in the Dockerfile 8 is that it is also needed for the PDF feature of enterprise edition. Since the 7.9 LTS is community only for Docker, we don't have both need.
> 
> Is there a specific use case that you can not achieve with the built-in feature; that is accomplished by the svg-badge plugin ?

the fact is, we have many many repositories (with older branches) where still have the link to the "old"  badges in some .md files, so by removing the plugin these links are broken...

for newer branches / repositories we already switched to the new feature (which is great).

@pierre-guillot-sonarsource: I'm complete with you that the built-in functionality should be used... but as already stated otherwise the compatibility simply seems to be broken, which is somehow sad...

simonbrandhof:
couldn't we think about a native support of the svg-badge plugin URLs? Was that already discussed?

pierre-guillot-sonarsource:
Hi, we will stick to our policy for the LTS : only fix security vulnerabilities, or blockers. As you have another solution for badges (the bundled one), i won't consider this as a blocker. Thanks for your time investing this issue. Keep in mind that you still have the possibility to extend this image with your own needs.

PR #378, opened 2020-03-16
SONAR-13183: add Snyk master branch and pull request configuration

michal-duda-sonarsource:


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,31 @@
+name: CI
+
+on: [pull_request]
```
jeremy-davis-sonarsource: Nitpicking: you don't need to wrap it in an array

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the Docker image
+        run: docker build -t "sonarqube:${{ matrix.tag }}" "${{ matrix.version }}"
+      - name: Run Snyk
+        uses: snyk/actions/docker@master
+        env:
+          SNYK_TOKEN: ${{secrets.SNYK_TOKEN}}
+        with:
+          command: test
```
jeremy-davis-sonarsource: Could this be conditional on the event? The two workflows could be merged, then (essentially the same?)

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,31 @@
+name: CI
```
jeremy-davis-sonarsource: Give it a different name?

.github/workflows/pull_request.yml
michal-duda-sonarsource: 👍

.github/workflows/pull_request.yml
michal-duda-sonarsource: I refactored it a bit, it's better now but since github actions don't support YAML anchors, there is a bit of duplication anyway

.github/workflows/pull_request.yml
michal-duda-sonarsource: 👍

.github/workflows/qa.yml
```diff
@@ -0,0 +1,80 @@
+name: QA
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    - cron: '40 15 * * *'
```
simonbrandhof: Why 40?

.github/workflows/qa.yml
```diff
@@ -0,0 +1,80 @@
...
+        if: github.event_name != 'pull_request'
+        uses: snyk/actions/docker@master
+        env:
+          SNYK_TOKEN: ${{secrets.SNYK_TOKEN}}
+        with:
+          command: monitor
+          image: sonarqube:${{ matrix.tag }}
+          args: --project-name=docker-sonarqube-${{ matrix.tag }} --file=${{ matrix.version }}/Dockerfile
+
+  snyk_test:
```
simonbrandhof: Snyk already runs on `master` in the job `build_and_test`, so to drop ambiguity I suggest to rename this job `snyk_test` to something like `snyk_on_pr`.

.github/workflows/qa.yml
michal-duda-sonarsource: I've set it to a time that it used to run on travis (from travis logs).

simonbrandhof: Ah ok :-) We don't control the scheduled time at Travis, so I suggest to keep things simple and use noon or midnight.

michal-duda-sonarsource: done

.github/workflows/qa.yml
michal-duda-sonarsource: Done.

PR #368, opened 2020-02-12
Linting, GPG signature verification, and Docker secrets

konstruktoid:
Basic linting of the Dockerfiles,
Adds GPG signature verification (closes #363),
Uses `--no-install-recommends` (closes #349),
Adds skeleton support for Docker secrets (closes #156)

```
$ docker images sonarqube
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
sonarqube           local-latest        c5e326b32f3f        18 minutes ago      570MB
sonarqube           developer-beta      cff979697db0        9 days ago          681MB
sonarqube           community-beta      9f896e5a7de4        9 days ago          575MB
sonarqube           latest              0956756d5b96        9 days ago          483MB
```

```
printf "sonarqube" | docker secret create sonarqube_jdbc_password -
docker service create --name sonarqube --secret sonarqube_jdbc_password -p 9000:9000 sonarqube
```

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

PR #379, opened 2020-03-22
SONAR-13182 switch to alpine

michal-duda-sonarsource:
- switch base image to alpine
- allow running the image as an arbitrary user
- support running the image in read-only mode (when using volumes)

### DIFF
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -1,46 +1,64 @@
-FROM openjdk:11-jre-slim
-
-RUN apt-get update \
-    && apt-get install -y curl unzip gnupg2 libfreetype6 libfontconfig1 \
-    && rm -rf /var/lib/apt/lists/*
-
-# Http port
-EXPOSE 9000
-
-RUN groupadd -r sonarqube && useradd -r -g sonarqube sonarqube
-
 ARG SONARQUBE_VERSION=8.2.0.32929
+
+FROM adoptopenjdk/openjdk11:alpine-jre as build
+ARG SONARQUBE_VERSION
 ARG SONARQUBE_ZIP_URL=https://binaries.sonarsource.com/Distribution/sonarqube/sonarqube-${SONARQUBE_VERSION}.zip
-ENV SONAR_VERSION=${SONARQUBE_VERSION} \
-    SONARQUBE_HOME=/opt/sonarqube
+ENV SONARQUBE_HOME=/opt/sonarqube \
+    SQ_DATA_DIR="/opt/sonarqube/data" \
+    SQ_EXTENSIONS_DIR="/opt/sonarqube/extensions" \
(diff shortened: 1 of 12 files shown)


simonbrandhof:
The commit should reference SONAR-13156 first. And could the changes related to SONAR-13182  be split into dedicated commits, or at least listed?

konstruktoid:
@michal-duda-sonarsource @pierre-guillot-sonarsource 
so I don't get the logic here, you got #368 from February that closes a bunch of issues and uses Alpine as base with no comments at all and then you open a PR that creates even more conflicts, and even goes against https://github.com/SonarSource/sonar-scanner-cli-docker/pull/36#issuecomment-601320716 regarding using Alpine.

Is there a point with these public repositories?
ref https://github.com/SonarSource/sonar-scanner-cli-docker/issues/41

michal-duda-sonarsource:
@konstruktoid the issue here is every new feature has to be internally discussed and validated before being merged for which we didn't have time when you opened the PR, everyone was busy working on non-Docker stuff. Your PR addresses too many issues at once and breaks backwards compatibility with previous image which might interfere with existing workflows. Internally people are opposed to supporting Docker Secrets too.

As for Alpine- I don't think there is conflict with what I've said in SonarSource/sonar-scanner-cli-docker#36. We wanted to evaluate other base images (and we did) but in the end we settled on Alpine (at least for now) for SonarQube. It might be different for `scanner-cli` since we probably have more options there (like the Distroless image, which we couldn't use here).

I know it's after the fact but for what it's worth we updated the README to clarify our approach to contributions (and raising issues/asking questions).

petemounce:
> @konstruktoid the issue here is every new feature has to be internally discussed and validated before being merged for which we didn't have time when you opened the PR, everyone was busy working on non-Docker stuff. Your PR addresses too many issues at once and breaks backwards compatibility with previous image which might interfere with existing workflows. Internally people are opposed to supporting Docker Secrets too.
> 
> As for Alpine- I don't think there is conflict with what I've said in [SonarSource/sonar-scanner-cli-docker#36](https://github.com/SonarSource/sonar-scanner-cli-docker/pull/36). We wanted to evaluate other base images (and we did) but in the end we settled on Alpine (at least for now) for SonarQube. It might be different for `scanner-cli` since we probably have more options there (like the Distroless image, which we couldn't use here).
> 
> I know it's after the fact but for what it's worth we updated the README to clarify our approach to contributions (and raising issues/asking questions).

@michal-duda-sonarsource would you mind creating a `CONTRIBUTING.md` that sets those expectations? github will then auto-link that inside of the UI when a PR is opened. https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

michal-duda-sonarsource:
@petemounce good idea 👍. Done.

PR #425, opened 2020-09-19
Dockerfile updated: Added RUN chmod +x for run.sh and sonar.sh

rjvrijn:
If run.sh and sonar.sh in your own environment doesn't have the execution attribute, building the new image will fail with a `permission denied` error. I had created both files manualy and forgot to set the -x permission. If set the execution attribute within de Dockerfile you dont have to worry about it anymore.


tobias-trabelsi-sonarsource:
Hi @rjvrijn ,
you are correct but this would introduce a new layer to the docker image and the permission flag is carried over to the build environment with git already, so i would reject these changes but thank you for taking the time to write this pullrequest (:
 
please had over to our community forum whenever you are facing issues with sonarqube and/or our docker images :+1:

PR #429, opened 2020-09-28
updated java and glibc

tobias-trabelsi-sonarsource:
used upstream reference: https://github.com/AdoptOpenJDK/openjdk-docker/blob/master/11/jdk/alpine/Dockerfile.hotspot.releases.slim
also updated glibc as announced here: https://sourceware.org/pipermail/libc-announce/2020/000029.html

### DIFF
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -7,39 +7,38 @@ ENV JAVA_VERSION="jdk-11.0.6+10" \
 #
 # glibc setup
 #
-RUN set -eux; \
-    apk add --no-cache --virtual .build-deps curl binutils; \
-    GLIBC_VER="2.31-r0"; \
-    ALPINE_GLIBC_REPO="https://github.com/sgerrand/alpine-pkg-glibc/releases/download"; \
-    GCC_LIBS_URL="https://archive.archlinux.org/packages/g/gcc-libs/gcc-libs-9.1.0-2-x86_64.pkg.tar.xz"; \
-    GCC_LIBS_SHA256="91dba90f3c20d32fcf7f1dbe91523653018aa0b8d2230b00f822f6722804cf08"; \
-    ZLIB_URL="https://archive.archlinux.org/packages/z/zlib/zlib-1%3A1.2.11-3-x86_64.pkg.tar.xz"; \
-    ZLIB_SHA256=17aede0b9f8baa789c5aa3f358fbf8c68a5f1228c5e6cba1a5dd34102ef4d4e5; \
-    curl -LfsS https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub -o /etc/apk/keys/sgerrand.rsa.pub; \
-    SGERRAND_RSA_SHA256="823b54589c93b02497f1ba4dc622eaef9c813e6b0f0ebbb2f771e32adf9f4ef2"; \
-    echo "${SGERRAND_RSA_SHA256} */etc/apk/keys/sgerrand.rsa.pub" | sha256sum -c -; \
-    curl -LfsS ${ALPINE_GLIBC_REPO}/${GLIBC_VER}/glibc-${GLIBC_VER}.apk > /tmp/glibc-${GLIBC_VER}.apk; \
-    apk add --no-cache /tmp/glibc-${GLIBC_VER}.apk; \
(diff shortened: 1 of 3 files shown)


8/community/Dockerfile
```diff
@@ -8,38 +8,38 @@ ENV JAVA_VERSION="jdk-11.0.6+10" \
...
     apk add --no-cache /tmp/glibc-i18n-${GLIBC_VER}.apk; \
     /usr/glibc-compat/bin/localedef --force --inputfile POSIX --charmap UTF-8 "$LANG" || true; \
     echo "export LANG=$LANG" > /etc/profile.d/locale.sh; \
-    curl -LfsS ${GCC_LIBS_URL} -o /tmp/gcc-libs.tar.xz; \
-    echo "${GCC_LIBS_SHA256} */tmp/gcc-libs.tar.xz" | sha256sum -c -; \
+    curl -LfsS ${GCC_LIBS_URL} -o /tmp/gcc-libs.tar.zst; \
+    echo "${GCC_LIBS_SHA256} */tmp/gcc-libs.tar.zst" | sha256sum -c - ; \
     mkdir /tmp/gcc; \
-    tar -xf /tmp/gcc-libs.tar.xz -C /tmp/gcc; \
+    zstd -d /tmp/gcc-libs.tar.zst --output-dir-flat /tmp; \
```
malena-ebert-sonarsource, line 32: With `tar -I zstd ...` the archive could be extracted in one line.

8/community/Dockerfile
```diff
@@ -48,20 +48,24 @@ RUN set -eux; \
     ARCH="$(apk --print-arch)"; \
     case "${ARCH}" in \
        aarch64|arm64) \
-         ESUM='7ed04ed9ed7271528e7f03490f1fd7dfbbc2d391414bd6fe4dd80ec3bad76d30'; \
-         BINARY_URL='https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.6%2B10/OpenJDK11U-jre_aarch64_linux_hotspot_11.0.6_10.tar.gz'; \
+         ESUM='fb27ea52ed901c14c9fe8ad2fc10b338b8cf47d6762571be1fe3fb7c426bab7c'; \
+         BINARY_URL='https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.8%2B10/OpenJDK11U-jdk_aarch64_linux_hotspot_11.0.8_10.tar.gz'; \
+         ;; \
+       armhf|armv7l) \
```
malena-ebert-sonarsource, line 54: Do we support arm? Or better, why it was not in this list before?

8/community/Dockerfile
tobias-trabelsi-sonarsource: the tar version in alpine does not support this flag. i think it's only available in GNU tar
```bash
tar -I zstd /tmp/gcc-libs.tar.zst --output-dir-flat /tmp/gcc
tar: unrecognized option: I
BusyBox v1.31.1 () multi-call binary.

Usage: tar c|x|t [-ZzJjahmvokO] [-f TARFILE] [-C DIR] [-T FILE] [-X FILE] [--exclude PATTERN]... [FILE]...

Create, extract, or list files from a tar file

        c       Create
        x       Extract
        t       List
        -f FILE Name of TARFILE ('-' for stdin/out)
        -C DIR  Change to DIR before operation
        -v      Verbose
        -O      Extract to stdout
        -m      Don't restore mtime
        -o      Don't restore user:group
        -k      Don't replace existing files
        -Z      (De)compress using compress
        -z      (De)compress using gzip
        -J      (De)compress using xz
        -j      (De)compress using bzip2
        -a      (De)compress using lzma
        -h      Follow symlinks
        -T FILE File with names to include
        -X FILE File with glob patterns to exclude
        --exclude PATTERN       Glob pattern to exclude
```

8/community/Dockerfile
tobias-trabelsi-sonarsource: not sure. maybe it wasn't available by adoptopenjdk? 
but in the end this does not matter as we only publish `linux/amd64` anyway: https://hub.docker.com/_/sonarqube?tab=tags

malena-ebert-sonarsource (approved):
LGFM

PR #435, opened 2020-10-12
Check presence of SONARQUBE_*

gbonnefille:
When SONARQUBE_* are not set, previous implementation pass -D options with empty values.
Doing this, any values set in the property file are ignored.

This patch avoid setting -D options when var env are not set.

- [ ] If the PR modifies or adds images, use the `run-tests.sh imagedir` command to verify the image works
- [ ] If the PR modifies or adds images, try to make sure the images run correctly on Linux

### DIFF
--- a/7/community/run.sh
+++ b/7/community/run.sh
@@ -21,12 +21,22 @@ do
     fi
 done < <(env)
 
+if [ -n "$SONARQUBE_JDBC_USERNAME" ]
+then
+    sq_opts+=("-Dsonar.jdbc.username='$SONARQUBE_JDBC_USERNAME'")
+fi
+if [ -n "$SONARQUBE_JDBC_PASSWORD" ]
+then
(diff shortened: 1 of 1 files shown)


tobias-trabelsi-sonarsource:
Hi @gbonnefille ,

thanks for your contribution. i could reproduce this and confirm that your fix will solve this. one small thing thou: could you move the if statements to line 16, so the order of arguments does not change?

tobias-trabelsi-sonarsource (changes requested):
Hi @gbonnefille ,

thanks for your contribution. i could reproduce this and confirm that your fix will solve this. one small thing thou: could you move the if statements to line 16, so the order of arguments does not change?

gbonnefille:
> Hi @gbonnefille ,
> 
> thanks for your contribution. i could reproduce this and confirm that your fix will solve this. one small thing thou: could you move the if statements to line 16, so the order of arguments does not change?

Done.

tobias-trabelsi-sonarsource (approved):
LGTM

PR #473, opened 2021-03-01
Support forward-compatible env vars in 7.9 LTS

usmonster:
The `SONARQUBE_*` environment variables used in older images have been
deprecated in favor of sticking with the `SONAR_*` equivalents (cf.
[MMF-1881](https://jira.sonarsource.com/browse/MMF-1881)). This deprecation was only applied to Dockerfiles for versions
8.x, which had the side effect of making it slightly harder for the Helm
chart to support both major versions out of the box.

This change allows the non-deprecated environment variables to be read
by the 7.9 LTS image, if present, without breaking existing behavior.

- [x] If the PR modifies or adds images, use the `run-tests.sh imagedir` command to verify the image works
- [x] If the PR modifies or adds images, try to make sure the images run correctly on Linux

### DIFF
--- a/7/community/Dockerfile
+++ b/7/community/Dockerfile
@@ -5,6 +5,11 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 ENV SONAR_VERSION=7.9.6 \
+    SONAR_HOME="" \
+    SONAR_JDBC_USERNAME="" \
+    SONAR_JDBC_PASSWORD="" \
+    SONAR_JDBC_URL="" \
+    # Deprecated
     SONARQUBE_HOME=/opt/sonarqube \
     SONARQUBE_JDBC_USERNAME=sonar \
     SONARQUBE_JDBC_PASSWORD=sonar \
--- a/7/community/run.sh
+++ b/7/community/run.sh
@@ -14,17 +14,17 @@ fi
 
 declare -a sq_opts
 
-if [ -n "$SONARQUBE_JDBC_USERNAME" ]
+if [ -n "$SONAR_JDBC_USERNAME" ] || [ -n "$SONARQUBE_JDBC_USERNAME" ]
 then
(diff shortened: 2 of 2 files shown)


tobias-trabelsi-sonarsource:
Hey @usmonster 

we are discussing your PR internally as changes to the LTS are a little spicy in general. we will come back to you after we have an alignment.
On a technical side i have tested your PR and it looks good to me, so currently there is no change needed from your side, just patience :)

usmonster:
Thanks for the feedback, @tobias-trabelsi-sonarsource!

If it helps the discussion, this isn't a new feature or functional change, but it only tiny improvement that helps align images & simplify the upgrade path for those still using the LTS image with the official chart.

Looking forward to the decision!

pierre-guillot-sonarsource:
ping @tobias-trabelsi-sonarsource :)

tobias-trabelsi-sonarsource:
Hi @usmonster ,

sadly the PM has blocked changes to the LTS as the general policy allows only blocker bugfixes and security patches to the LTS. We will a new LTS release soon that will replace 7.9 with 8.10 (the new LTS version of sonarqube) and this version will include the current environment variable support, so hopefully you can work with that when it's released.

thank you for the contribution and the good idea you brought to the table

usmonster:
Hi @tobias-trabelsi-sonarsource. Thanks for the transparency, and I completely understand the policy.

In any case, here's an effective workaround we use in our HelmRelease values, in case anyone else wants to use the 7.x LTS with more recent versions of the official chart:

```yaml
   # Use SonarQube < 8 environment variables keys (see https://github.com/Oteemo/charts/issues/262 & https://github.com/SonarSource/docker-sonarqube/pull/473)
    env:
    - name: SONARQUBE_JDBC_USERNAME
      value: sonar
    - name: SONARQUBE_JDBC_PASSWORD
      valueFrom:
        secretKeyRef:
          key: postgresql-password
          name: sonarqube-pg
    - name: SONARQUBE_JDBC_URL
      value: jdbc:postgresql://X.X.X.X:5432/sonar
```

PR #515, opened 2021-11-25
Multiarch support

stevefan1999-personal:
This simple PR should add multiplatform support for running SonarQube on Raspberry Pi and Oracle Cloud ARM and Graviton2 and such


tobias-trabelsi-sonarsource:
Hi @stevefan1999-personal ,

thanks for the contribution. enabling arm builds is something that is currently discussed internally but there is no ETA on that. sadly the changes that you are suggesting is not the way to go as our images are part of the official library of the docker hub, so the test is not the only thing that needs to be adapted here. apart from that, the QA is failing because you removed a needed environment variable during build time.

PR #539, opened 2022-02-24
Fix running on M1 Mac

jafri:
Without making bin files executable, docker run fails on M1 mac

### DIFF
--- a/9/community/Dockerfile
+++ b/9/community/Dockerfile
@@ -51,6 +51,8 @@ RUN set -eux; \
     apk del --purge build-dependencies;
 
 COPY --chown=sonarqube:sonarqube run.sh sonar.sh ${SONARQUBE_HOME}/bin/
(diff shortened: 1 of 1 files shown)


tobias-trabelsi-sonarsource:
Hi @jafri ,

these scripts do already have the executable flag and do not need them again. this is only needed if you clone this repo on a filesystem that can not handle unix file permissions and try to build from there. 
```
ttrabelsi@verdandi github/docker-sonarqube/9/community [ git: master  ] > ls -alh
Permissions Size User      Date Modified Name
.rw-r--r--  2.7k ttrabelsi 25 Feb 08:16  Dockerfile
.rwxr-xr-x  2.2k ttrabelsi  8 Oct  2021  run.sh
.rwxr-xr-x   110 ttrabelsi  2 Jul  2021  sonar.sh
```

PR #545, opened 2022-04-01
REL-2214 Update to 9.4

wouter-admiraal-sonarsource:


8/community/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.13
+FROM alpine:3.13.9
```
pierre-guillot-sonarsource, line 1: This seems weird, as [3.13](https://github.com/alpinelinux/docker-alpine/blob/ed36cc54c3083ad4f1a84b2c8f63431cca2768cb/x86_64/Dockerfile) and [3.13.9](https://github.com/alpinelinux/docker-alpine/blob/ed36cc54c3083ad4f1a84b2c8f63431cca2768cb/x86_64/Dockerfile) are actually the same version

pierre-guillot-sonarsource (approved):
I'm quite worried about the fact that we have to force the patch version, which should not be needed.

I'm approving the PR in case we can't figure out a way to clear the cache

8/community/Dockerfile
wouter-admiraal-sonarsource: Let me try to remove the commit, and see if it busts the cache. Perhaps it will work. Still, weird that we run into this issue.

wouter-admiraal-sonarsource:
For future ref: I didn't manage to clear the cache. To unblock the current release, I'm fixing the base image versions to a specific patch version. We can always revisit this in the future if needed.

PR #562, opened 2022-07-31
add permission and set to rootless for community Dockerfile

hanhsuan:
The permission of run.sh and sonar.sh might be different than default.
Force to change the permission in Dockerfile could be more stable.
At the sametime, change the default user to sonarqube should be better.

After build and run on Linux and Kubernetes, everything is well.

### DIFF
--- a/8/community/Dockerfile
+++ b/8/community/Dockerfile
@@ -129,7 +129,12 @@ RUN set -eux; \
 
 COPY --chown=sonarqube:sonarqube run.sh sonar.sh ${SONARQUBE_HOME}/bin/
 
+RUN chmod 700 ${SONARQUBE_HOME}/bin/run.sh && chmod 700 ${SONARQUBE_HOME}/bin/sonar.sh
+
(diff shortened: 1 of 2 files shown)


wouter-admiraal-sonarsource:
Hi @hanhsuan ,

Thanks for taking the time to submit a PR. Unfortunately, I cannot accept it.

* Changing the mod is redundant, and unnecessary. We have a QA that will make sure we have no regressions here, so although I understand the reasoning, we don't need this.
* Changing the `USER` is a breaking change. For people upgrading from 9.5 to an upcoming version (like the soon-to-be-released 9.6), this will have unexpected side effects. Although I agree that setting `USER` is likely a better idea than letting the container run as root by default, it's not a change we can make lightly. Right now, the image's semantic versioning is tied to SonarQube itself. As this is a breaking change (i.e., requires a MAJOR version increment), we would probably consider changing `USER` only when 10.0 comes out (start of 2023).

For these reasons, I will close this PR without accepting it. Sorry.

Best regards.

PR #566, opened 2022-08-15
Build without cache

wouter-admiraal-sonarsource:


guillaume-peoch-sonarsource (approved):
LGTM

nixel2007:
@wouter-admiraal-sonarsource --pull forces docker to pull base image, but cache is still used. To build without cache you should add --no-cache flag

wouter-admiraal-sonarsource:
I was only concerned with the base image 🙂. Our QA kept failing because Trivy wasn't picking up the latest base image.

PR #571, opened 2022-10-10
Replace status badge

wouter-admiraal-sonarsource:
Let's replace the status badge. We're not using Travis anymore, but use GitHub Actions.


carminevassallo (approved):
LGTM 😃

PR #579, opened 2022-11-18
BUILD-2211 enable access to sonardocker account in vault

tomverin:


.cirrus.yml
```diff
@@ -1,3 +1,12 @@
+env:
+  # see https://github.com/SonarSource/re-terraform-aws-vault/blob/master/orders/bubble-cfamily.yaml
+  CIRRUS_VAULT_URL: https://vault.sonar.build:8200
```
carminevassallo, line 3: This repository is public. Is it still fine to have this URL unmasked?

carminevassallo (commented):
Just a comment on the vault URL. I've also set the PR to be merged on my feature branch, as the `.cirrus.yml` is still under construction :)

.cirrus.yml
tomverin: yes it's like that for all public repos

PR #592, opened 2022-12-21
DOCKER-134 ARM64 support and eclipse-temurin as base image

carminevassallo:
We need to build our arm64-compatible image in a CI environment and make sure that we can test it properly.


9/community/Dockerfile
```diff
@@ -7,11 +7,10 @@ ENV LANG='en_US.UTF-8' \
 #
 # SonarQube setup
 #
-ARG JAVA_VERSION=17
+ARG JAVA_VERSION=17 #JAVA_VERSION argument kept for retro-compatibility: do we really need it?
```
carminevassallo, line 10: Can we fix this comment before merging?

jCOTINEAU (approved):
lgtm

9/community/Dockerfile
carminevassallo: Ok then I'm gonna remove it.

PR #595, opened 2023-01-13
[NO-JIRA] Remove unnecessary properties setup

carminevassallo:
We set several properties which are not mandatory. This PR has the goal of cleaning them up.

### DIFF
--- a/9/datacenter/search/run.sh
+++ b/9/datacenter/search/run.sh
@@ -25,25 +25,10 @@ fi
 if [[ "$1" = '/opt/sonarqube/bin/sonar.sh' ]]; then
 
     #
-    # Change log path to ensure every search node can write in their own directory
-    # This resolves a cluttered log on docker-compose with scale > 1
-    #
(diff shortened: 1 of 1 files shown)


jCOTINEAU (approved):
lgtm.
We need to remember to change the docker-compose example related to the dce to add that path inside of it and in the documentation.

jacek-poreda-sonarsource (approved):
LGTM 👍

carminevassallo:
I think @jCOTINEAU has a very good point. I'll update the examples in the context of this PR already.

carminevassallo:
Given that the app nodes are identical replicas part of a Deployment, at the moment we can't avoid the cluttered-logs issue in these nodes. The problem is not affecting search nodes instead.

Therefore, we are gonna remove the cluttered-logs patch in the search nodes, but not in the app nodes. This would instead require some changes within the application.

jCOTINEAU (approved):
lgtm !

PR #598, opened 2023-02-02
DOCKER-159 address docker hub maintainers comments

jCOTINEAU:
- add entrypoint for compliance with [official-images](https://github.com/docker-library/official-images/tree/05f539b9a91128e2238f584d6dc4f0a81a8e1a9f#consistency)
- rework the way we add user to prevent belonging to `no_group`
- removing unneeded apt list after install
- revert changes made in #553, no longer required after tests on windows

### DIFF
--- a/9/community/Dockerfile
+++ b/9/community/Dockerfile
@@ -18,9 +18,8 @@ ENV JAVA_HOME='/opt/java/openjdk' \
     SQ_TEMP_DIR="/opt/sonarqube/temp"
 
 RUN set -eux; \
-    adduser --system --uid 1000 sonarqube; \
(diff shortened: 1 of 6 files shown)


carminevassallo (approved):
The changes look great!

We should just mention (maybe in the PR description) that such PR reverts the changes made [here](https://github.com/SonarSource/docker-sonarqube/pull/553), as they are not needed anymore after changing the base image. (We tested it on @antoine-vinot-sonarsource's machine)

jCOTINEAU:
I absolutely agree, i will mark that in the PR and in the commit description

PR #607, opened 2023-03-22
DOCKER-166 Update release process

jCOTINEAU:


release.md
```diff
@@ -1,5 +1,9 @@
 # Releasing
 
+Docker image release cycle and SonarQube product
+---
+We consider the **docker image** as part of the SonarQube **product**
```
carminevassallo, line 5: I would just add a small sentence:

We consider the **docker image** as part of the SonarQube **product**. Therefore, it follows the same release process.

PR #616, opened 2023-04-19
SONAR-18978 Fix the dependency between app and search nodes

carminevassallo:


jCOTINEAU (approved):
lgtm, tested on my side, everything works fine, great added value !

PR #628, opened 2023-06-14
SONAR-19539 add DOCKER_RUNNING env variable to our Dockerfiles

davividal:


jCOTINEAU (approved):
lgtm ! 

side question, because the telemtry code will not be ported to SQ 9.9, it might not be needed to change the 9th folder in the end ?

PR #634, opened 2023-08-29
SQDEVOPS-160 improve-release-doc-related-to-forks

jCOTINEAU:


release.md
```diff
@@ -43,17 +43,23 @@ For more and up to date documentation, see https://github.com/docker-library/doc
 Update Docker Hub's SonarQube images
 -----------------------
 
-Update the SonarSource [fork](https://github.com/SonarSource/official-images) of the [official-images](https://github.com/docker-library/official-images) to ensure that the `sonarqube` library is the latest version.
+In order to update the Docker Hub images, a Pull Request must be created on the [official-images](https://github.com/docker-library/official-images) repository.
 
-Create a feature branch on the company fork:
+To do so you must use a fork, you can either:
```
carminevassallo, line 48: If the SonarSource fork does not work as expected, let's just suggest the solution that works. We could even delete that fork if it's useless...

carminevassallo (changes requested):
Just a few minor changes. Let's just suggest what works.

release.md
jCOTINEAU: i agree, it does not add any value to keep that fork, rephrasing in that direction.

Should we still mention it existed at some point ? to give a bit of context ? Not sure that would be needed

carminevassallo (approved):

PR #649, opened 2023-10-02
SONAR-20590 - Adds retry mechanism for public and private scan tasks

davividal:


carminevassallo (changes requested):
Hi @davividal,

great PR, the script is super easy to read.

I think we can still optimize it a bit. WDYT?

1.  The two scripts are almost equal. We can easily refactor and extract some variables to have only one script for both CI steps.
2. The issue was only about docker pull[ing] from DockerHub. I propose we remove the retry of the scan on Mend, as we never face issues there.

jCOTINEAU (approved):
lgtm

carminevassallo (approved):

PR #651, opened 2023-10-25
SONAR-20857 Keeps curl to be used by Helm Chart

davividal:


jCOTINEAU (approved):
LGTM. 
As discussed keeping curl will not drastically increase the size and does not increase the attack vector.

PR #653, opened 2023-11-02
SONAR-20918 Fix security hotspot docker:S6500

carminevassallo:


10/community/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN set -eux; \
     groupadd --system --gid 1000 sonarqube; \
     useradd --system --uid 1000 --gid sonarqube sonarqube; \
     apt-get update; \
-    apt-get install -y gnupg unzip curl bash fonts-dejavu; \
+    apt-get --no-install-recommends install -y gnupg unzip curl bash fonts-dejavu; \
```
davividal, line 27: Pet peeve: `apt-get --no-install-recommends -y install ...`

davividal: @carminevassallo, it is not a question of bad style.
The [man page for apt-get](https://linux.die.net/man/8/apt-get) shows that the expected order is `apt-get [options] command [pkgs]`. Historically, it works both ways, but you shouldn't rely on it.

davividal (approved):
LGTM!

PR #660, opened 2023-11-20
SONAR-21071 Update the scan policy for SQ Docker images

carminevassallo:


.cirrus.yml
```diff
@@ -64,7 +65,7 @@ private_scan_task:
     WS_WSS_URL: https://saas-eu.whitesourcesoftware.com/agent
     matrix:
       - tag: $CURRENT_LTS_VERSION-datacenter-app
-      - tag: $CURRENT_VERSION-datacenter-app
+      - tag: $NEXT_VERSION-datacenter-app
```
davividal, line 68: Quick question: the ticket mentions that we test the latest LTS, the latest version, and the next version. But here, the latest version (`$CURRENT_VERSION`) is not being tested anymore?

davividal (changes requested):

.cirrus.yml
carminevassallo: In the private scan, we test the latest LTS and the next version. In the public scan below, we test the latest (released) LTS and the latest (released) version. The difference among the two scans is that the private scan is run on every build, as it scans images which are under active (or semi-active in case of LTS) development. The public scan is only triggered during the nightly and includes released images only. Does this clarify your doubts?

davividal (approved):
LGTM!

PR #680, opened 2024-04-11
SONAR-19108 fix docker compose example app replica issue

jCOTINEAU:


carminevassallo (approved):
Hey Jeremy, the PR looks good to me! Great that we fix those examples!

I was just suggesting that we revisit memory limit and reservation (I got resource-related errors using the settings we had before), and we could bump the postgresql version we use. Feel free to squash my suggested changes...

PR #702, opened 2024-08-22
Update Dockerfile to patch base OS

dtm966:
without patching the base OS Nessus scans show critical vulnerabilities.

- [ ] If the PR modifies or adds images, use the `run-tests.sh imagedir` command to verify the image works
- [ ] If the PR modifies or adds images, try to make sure the images run correctly on Linux

### DIFF
--- a/10/community/Dockerfile
+++ b/10/community/Dockerfile
@@ -31,6 +31,7 @@ ENV ES_TMPDIR="${SQ_TEMP_DIR}"
 RUN set -eux; \
     useradd --system --uid 1000 --gid 0 sonarqube; \
     apt-get update; \
+    apt-get upgrade -y; \
     apt-get --no-install-recommends -y install \
         bash \
         curl \


jCOTINEAU:
hello @dtm966, do you mean critical vulnerabilities on the dockerHub images?
Because those are rebuilt automatically when the base image change, and should always have the latest available fixes.

On top of that, vulnerabilities in docker needs to be looked at in depth, because we use only a very small portion of what is inside the image, for example we had multiple critical severity on the web server of ubuntu, which is never ever used when spinning the SonarQube java process.

dtm966:
Yes- there's a breakdown somewhere so that older vulnerable versions of libs are found in the container - my particular issue is with libcurl. While I do understand the focus should be the project codebase, the artifacts left in the base container can affect overall system security. Doing a apt upgrade in the docker build ensures that the container is as secure as possible no matter what is happening upstream of you in the process

jCOTINEAU:
Thanks a lot for your feedback, we will discuss this internally with the team an keep you posted on this.

jCOTINEAU:
Hello again, @dtm966. After an internal discussion, we think it does not make sense to include the upgrade in our final Docker image.

The reasoning is that we trust and depend on Eclipse Temurin; in the chain of responsibility, we will control and do sticky-versioning of the packages we install and leave the others as is. We treat the base image/os/packages kind of as a black box.

We know it induces some noise and that it's a fair question, we will take a deeper look at how eclipse-temurin and ubutu do it and maybe ask for an improvement on this side.

Will close the PR for now, thanks a lot again for triggering such discussions

PR #704, opened 2024-08-27
NO-JIRA fix next-issues regarding malformed volumes

jCOTINEAU:


carminevassallo (approved):
LGTM! Just a minor comment below.

I do not think we have a formal process for it but we should change the commit message to something that can be understood externally. How about "NO-JIRA Fix docker:S7030 issues"? People will be able to check the actual issues in the docs...

PR #707, opened 2024-09-30
REL-3303 address reviewer comments removing no-op apt clean

jCOTINEAU:
As stated by docker official images maintainer, apt clean is not needed because the eclipse temurin base image uses hooks to every dpkg / apt command to clean afterward. Hence apt clean here is a no-op entry


matteo-mara-sonarsource (approved):
LGTM!

PR #711, opened 2024-10-16
SONAR-23408 Run FOSSA analysis

matteo-mara-sonarsource:


.cirrus/tasks.yml
```diff
@@ -153,6 +153,24 @@ public_scan_task:
...
+  ec2_instance:
+    <<: *VM_TEMPLATE
+  setup_script:
+    - curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install-latest.sh | bash
+  scan_script:
+    - echo "Scan the ${PUBLIC_IMAGE_NAME}:${tag} image"
+    - docker pull "${PUBLIC_IMAGE_NAME}:${tag}"
+    - .cirrus/pull.sh ${PUBLIC_IMAGE_NAME} ${tag} ${platform}
+    - fossa container analyze -p "docker-sonarqube" -b "${tag}"
+  depends_on: multi_arch_build
```
carminevassallo, line 172: This is a scan of the public image. I'd suggest we rename the task's name accordingly and remove the dependence on `multi_arch_build` as we don't need to build the image to run it.

carminevassallo (changes requested):

.cirrus/tasks.yml
matteo-mara-sonarsource: I was not sure if it was better to scan the STAGING or the PUBLIC image.
I am noticing now that I missed the first echo.
Moving everything to PUBLIC and dropping the dependency on multi_arch_build

carminevassallo (approved):
LGTM 💯

PR #741, opened 2025-01-27
SONAR-24186 improve file ownership

jCOTINEAU:
[SONAR-24186](https://sonarsource.atlassian.net/browse/SONAR-24186)

We identified that we were not covering all options for runAsUser and runAsGroup in kubernetes context.

By adding ownership of files to sonarqube user we are compatible with both those workflow:

the current openshift behavior which is anyUid and gid=0
More classic use case were people rely on UID of the image (1000 in our case) but with gid!=0

Those two workflow will work as permissions are assigned to both owner and group.

[SONAR-24186]: https://sonarsource.atlassian.net/browse/SONAR-24186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


davividal (approved):
LGTM!

PR #745, opened 2025-02-05
NO-JIRA address docker maintainer feedback

jCOTINEAU:


2025/datacenter/app/Dockerfile
```diff
@@ -59,14 +59,16 @@ RUN set -eux; \
     rm sonarqube.zip*; \
     rm -rf ${SONARQUBE_HOME}/bin/*; \
     ln -s "${SONARQUBE_HOME}/lib/sonar-application-${SONARQUBE_VERSION}.jar" "${SONARQUBE_HOME}/lib/sonarqube.jar"; \
+    chown -R sonarqube:root ${SONARQUBE_HOME}; \
+    chown -R sonarqube:root "${SQ_DATA_DIR}" "${SQ_EXTENSIONS_DIR}" "${SQ_LOGS_DIR}" "${SQ_TEMP_DIR}"; \
```
carminevassallo, line 63: Why can't we merge those two in one line?

carminevassallo (changes requested):

2025/datacenter/app/Dockerfile
jCOTINEAU: we could indeed, my ocd like having them split because they are mapped to two different lines below. but i dont have strong preferances

carminevassallo: We can reduce the size of the file, I maybe merging the chown and chmod on one line could improve readability overall. Let's do this in a hardening though, it should not block the release.

carminevassallo (approved):

davividal (approved):
LGTM!

PR #751, opened 2025-02-21
Remove unsupported DCE Compose example

waynekhan:
Hello,

Per the discourse, I'm proposing to remove this DCE example altogether. As implemented, all 5 containers will run on the same Docker host.


jCOTINEAU:
Well those examples are not meant for production but for people to understand the DCE logic as well as testing locally.

This has proven quite useful both for sonarsourcers and users, so i speak i, think we should keep it.

waynekhan:
Thanks @jCOTINEAU, let's close and continue internally.

PR #776, opened 2025-06-06
OPE-803 define new repo structure

jCOTINEAU:


carminevassallo (commented):
This is an awesome improvement of our CI process! Big Kudos 👏 

Apart from suggesting to convert the ticket to `SONAR` and update the required tasks, I'm challenging this part: as of now, when we start a new release we set the the current version to be the next release in both SonarQube Server and its helm chart. I know that the changes on the dockerfiles are not necessarily big, but shall we set the CI to be aligned with what we do on the other repos? I.e., setting the next release as the current version.. WDYT?

jCOTINEAU:
> This is an awesome improvement of our CI process! Big Kudos 👏
> 
> Apart from suggesting to convert the ticket to `SONAR` and update the required tasks, I'm challenging this part: as of now, when we start a new release we set the the current version to be the next release in both SonarQube Server and its helm chart. I know that the changes on the dockerfiles are not necessarily big, but shall we set the CI to be aligned with what we do on the other repos? I.e., setting the next release as the current version.. WDYT?

Indeed, it would be great to streamline all release processs to have the same logic, and it will not cause any CI issues to have this.

PR #781, opened 2025-06-17
DRAFT SONAR-25280 automate generation of official-images release

jCOTINEAU:
[SONAR-25280](https://sonarsource.atlassian.net/browse/SONAR-25280)

[SONAR-25280]: https://sonarsource.atlassian.net/browse/SONAR-25280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker-official-images/internal/build/build.go
```diff
@@ -0,0 +1,281 @@
...
+
+		// Extract SONARQUBE_VERSION from the fetched content.
+		reader := strings.NewReader(content)
+		version, err := ExtractSonarQubeVersion(reader)
+		if err != nil {
+			return nil, fmt.Errorf("failed to extract SONARQUBE_VERSION from %s/%s: %w", branchOrCommit, relPath, err)
+		}
+
+        // Extract EditionType here
+        editionType, err := GetEditionTypeFromPath(relPath)
```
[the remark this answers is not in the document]
davividal: Also, the analyzer seems to be missing this somehow? The indentation shown here is different from the one in SonarQube:

https://next.sonarqube.com/sonarqube/code?id=SonarSource_docker-sonarqube_AYcnOvlJTpBOcQuGEdI5&pullRequest=781&selected=SonarSource_docker-sonarqube_AYcnOvlJTpBOcQuGEdI5%3Adocker-official-images%2Finternal%2Fbuild%2Fbuild.go&line=149

Maybe we need a dogfood thread on this.

This file is mixing tabs and spaces.

docker-official-images/internal/config/config_test.go
```diff
@@ -0,0 +1,344 @@
...
+			"type": "commercialEditions",
+			"commitSha": "deadbeef"
+		}
+	]`
+
+	// Write the JSON content to the temporary file
+	err = os.WriteFile(configFilePath, []byte(jsonContent), 0644) // 0644 is standard file permissions
+	if err != nil {
+		t.Fatalf("Failed to write temp config file: %v", err)
+	}
```
davividal, line 300: If you really want to dynamically create config tests, this can be a helper function: given a string, it returns the path of the temp file created.

I would avoid doing that, and I would split the `config.ParseConfigFile` instead, because I (speak I) am not interested in the create-file-open-file-read-file logic: I assume that it works. I am interested only if given `{"this":"json"}`, `config.ParseConfigFile` will fail because it isn't a valid config file.

docker-official-images/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+		metadataList, err := build.GetBuildMetadataFromConfig(cfg, gitFetcher)
+		if err != nil {
+			log.Fatalf("Error processing active config for branch %q: %v", cfg.Branch, err)
+		}
+		allBuildMetadata = append(allBuildMetadata, metadataList...)
+	}
+
+	fmt.Printf("Successfully processed %d image build metadata entries.\n", len(allBuildMetadata))
+	// fmt.Println("--- Generated Build Metadata ---")
```
davividal, line 36: This seems like debug information.
Why not using [logging](https://www.honeybadger.io/blog/golang-logging/)? We can use logrus and debug level. Or remove these comments. :)

docker-official-images/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// 1. Read and parse the activeVersions JSON file
+	activeConfigs, err := config.ParseConfigFile(configFilePath)
+	if err != nil {
+		log.Fatalf("Error reading or parsing active versions config: %v", err)
+	}
+
+    gitFetcher := build.NewGitFetcher() // Operates on current Git repo
+
+	// 3. Process each activeVersion to retrieve ImageBuildMetadata
```
davividal, line 24: Where is step 2? :)

docker-official-images/main.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	fmt.Printf("Successfully processed %d image build metadata entries.\n", len(allBuildMetadata))
+	// fmt.Println("--- Generated Build Metadata ---")
+	// for _, md := range allBuildMetadata {
+	// 	fmt.Printf("Branch: %s, Edition: %s, Version: %s, Directory: %s, Tags: %v, GitCommit: %s\n",
+	// 		md.Branch, md.EditionType, md.Version, md.ImageDirectory, md.Tags, md.GitCommit)
+	// }
+
+	templateFilePath := "official_images.tmpl"
```
davividal, line 42: Can we split this `main` into multiple functions so it is easier to test?

docker-official-images/main_test.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"log"
+	"os"
+	"testing"
+)
+
+// TestMainFunction provides a basic test for the main function.
+func TestMainFunction(t *testing.T) {
```
davividal, line 10: I think that one very important test to make here is to get the official images file from the official repo ([sonarqube](https://github.com/docker-library/official-images/blob/3501d02bf2a2551f950063778437206f347207f9/library/sonarqube)) and use it as a fixture.
Then, create the necessary test data (e.g. a test `active_versions.json`) that will generate that file, and compare what we generate with the fixture.

Currently, the official file has a slightly different structure (global GitCommit, duplicated Architectures for all the images, etc), but we can use that as a base for a fixture.

WDYT?

sonar-project.properties
```diff
@@ -1,3 +1,5 @@
 sonar.projectKey=SonarSource_docker-sonarqube_AYcnOvlJTpBOcQuGEdI5
-sonar.sources=9/,commercial-editions/,community-build/
+sonar.sources=9/,commercial-editions/,community-build/,docker-official-images/
 sonar.organization=sonarsource
+sonar.go.coverage.reportPaths=docker-official-images/coverage.out
+sonar.exclusions=docker-official-images/**/*_test.go
```
davividal, line 5: Why excluding the tests from the analysis?

davividal (changes requested):
I made some comments, but the most important thing is to run a `go fmt` in all the files first. I might have missed some things due to the inconsistent indentation.

docker-official-images/main.go
jCOTINEAU: i dont think it make sense, this main function is like a CI script, it compile other tested functions.

sonar-project.properties
jCOTINEAU: right, i used the wrong property, there is a dedicated sonar.tests

docker-official-images/main_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	main()
```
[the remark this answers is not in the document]
jCOTINEAU: main is ran in only one test, and any way the generated file override the other for idempotence

coverage.out
```diff
@@ -0,0 +1 @@
+mode: set
```
davividal, line 1: Do we want to commit this file?

davividal (changes requested):
There are some other reviews still unresolved.

As soon as we resolve these, I'm happy with this PR. :)

docker-official-images/internal/build/build.go
```diff
@@ -0,0 +1,281 @@
...
+func NewGitFetcher() *GitFetcher {
+	return &GitFetcher{}
+}
+
+// Fetch retrieves the content of a file from the current local Git repository
+// at a specific branch/commit without checking out the local directory.
+func (f *GitFetcher) Fetch(branchOrCommit, relativePath string) (string, error) {
+	// Use 'git show' to get the content of the file at the specific branch/commit.
+	// Format: <ref>:<path_to_file>
+	cmd := exec.Command("git", "show", fmt.Sprintf("%s:%s", branchOrCommit, relativePath))
```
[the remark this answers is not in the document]
davividal: I have no opinion on this. :)

I thought that the error handling could be easier but, as you said, we are well covered with tests and this is a well-controlled environment.

davividal (approved):
LGTM!

jCOTINEAU (commented):
Looks like there is a bug.
The PR cannot be merged even if all checks are passing, but the review one seems to be missing ?

Edit: writing this comment solved the issue

PR #926, opened 2026-09-07
SONAR-31880 Extend Trivy Scan to all the images and fail the build on medium or higher CVSS band

jimil09:
----
## Summary by Gitar

- **Security Scanning:**
    - Added Trivy vulnerability scanning across all images with medium or higher CVSS band policy enforcement
    - Added `.trivyignore.yaml` for governed and time-expired CVE suppressions
- **CI/CD Workflows:**
    - Created `evaluate-trivy-report.py` script with comprehensive unit tests for CVSS scoring and policy checks

<sub>This will update automatically on new commits.</sub>


jimil09:
Fixed in 250e8c1: replaced the invalid nested `report:` mapping in `.github/trivy.yaml` with the correct top-level `format:` key, and removed the five `TRIVY_REPORT: 'all'` workaround lines that were masking the underlying config bug. Good catch — thanks.

hatem-amairi-sonarsource:
Hello @jimil09 ,
Can you please check the behavior on vulnerabilities classified as `UNKNOWN` ?
I think we should block on that and to bring visibility to the operator who then takes a decision (ignore them via `.trivyignore`)

jimil09:
Good catch @hatem-amairi-sonarsource — agreed, fixed in 055bc4b. UNKNOWN (no CVSS score anywhere and no usable Trivy severity, i.e. zero information) now also fails the gate, forcing the same governed `.trivyignore.yaml` decision as any other Medium+ finding. Kept it distinct from a real CVSS score of 0.0 (NONE band), which is informed data saying "no impact" and still passes.

.github/workflows/push_and_pr.yml
```diff
@@ -309,7 +416,7 @@ jobs:
         id: alls-green
         with:
           jobs: ${{ toJSON(needs) }}
-          allowed-skips: 'load-config,build-commercial-editions,build-community-build,test-commercial-editions,test-community-build,trivy-scan-datacenter-app,gcp-build-staging-app,gcp-build-staging-search'
+          allowed-skips: 'load-config,build-commercial-editions,build-community-build,test-commercial-editions,test-community-build,gcp-build-staging-app,gcp-build-staging-search'
```
hatem-amairi-sonarsource: Both new scan jobs were dropped from allowed-skips, but their `if:` still carries
the editions filter. On a workflow_dispatch run with editions: community-only,
trivy-scan-commercial-editions skips, and alls-green v1.2.2 requires
result == 'success' for any job not listed in allowed-skips, so QA fails. The
same happens in reverse for commercial-only.

I think the premise behind ticket item 4 may not hold. Reading the pinned action
source, allowed-skips accepts only {'skipped', 'success'} - a job in that list
that *fails* still fails the gate. So the old trivy-scan-datacenter-app entry
was never suppressing enforcement; removing it doesn't strengthen the gate, and
it does break single-edition dispatch runs. The four sibling jobs using the same
editions-skip pattern (build-commercial-editions, build-community-build,
test-commercial-editions, test-community-build) are all still listed.

Suggestion: add both scan jobs back to allowed-skips.

.github/workflows/scripts/evaluate-trivy-report.py
```diff
@@ -0,0 +1,264 @@
...
+    rounded = math.ceil(score)
+    if rounded <= 0:
+        return "NONE"
+    if rounded <= 3:
+        return "LOW"
+    if rounded <= 6:
+        return "MEDIUM"
+    if rounded <= 8:
+        return "HIGH"
+    return "CRITICAL"
```
hatem-amairi-sonarsource, line 104: math.ceil(score) runs before the 0/3/6/8 cutoffs, which shifts every
non-integer score up one band:

    score      script band    CVSS v3.1 band
    3.1, 3.9   MEDIUM         LOW
    6.9        HIGH           MEDIUM
    8.9        CRITICAL       HIGH

Only the first row changes pass/fail, and it does so in the costly direction:
every fixable OS CVE scoring 3.1-3.9 is Low by the CVSS definition yet fails the
build, and each one then needs its own governed .trivyignore.yaml entry.

"Round up, never down" is already satisfied by banding the raw score against the
real cutoffs (3.9 / 6.9 / 8.9); the current code re-scales the band boundaries
instead. The tests encode the skew deliberately, so this reads as a deliberate
choice. Worth confirming with the ticket owner before anything changes.

.github/workflows/scripts/evaluate-trivy-report.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def _best_field(scores: dict[str, Any]) -> Optional[tuple[float, str]]:
+    """Return (score, version_label) for the highest-precedence populated field."""
+    for cvss_field, version in CVSS_VERSION_FIELDS:
+        value = scores.get(cvss_field)
+        if value:
```
hatem-amairi-sonarsource: `if value:` makes a genuine CVSS score of 0.0 falsy, so it reads as "no score".
Two consequences, both of which contradict the script's own docstring:

  - {"nvd": {"V3Score": 0.0}} with Severity: MEDIUM falls through to
    Trivy.Severity and bands as MEDIUM, so it fails. That is exactly the case
    the NONE-passes comment on lines 53-57 says should pass.

  - {"nvd": {"V3Score": 0.0}, "ghsa": {"V3Score": 7.0}} returns CVSS.ghsa.v3,
    which breaks the NVD-first precedence the ticket specifies.

Both verified by running the module directly. `if value is not None:` fixes both.

.github/workflows/scripts/evaluate-trivy-report.py
```diff
@@ -0,0 +1,264 @@
...
+        score, version = nvd_best
+        return score, f"CVSS.nvd.{version}"
+
+    best: Optional[tuple[float, str, str]] = None  # (score, source, version)
+    for source, scores in cvss.items():
+        if source == NVD_SOURCE or not scores:
+            continue
+        for cvss_field, version in CVSS_VERSION_FIELDS:
+            value = scores.get(cvss_field)
+            if not value:
```
hatem-amairi-sonarsource: Same issue here

.github/workflows/gcp-build-template.yml
```diff
@@ -85,6 +85,28 @@ jobs:
...
+          format: 'json'
+          output: 'trivy-release-datacenter-${{ inputs.component_type }}.json'
+          vuln-type: 'os'
+          trivy-config: '.github/trivy.yaml'
+          trivyignores: './.trivyignore.yaml'
+          ignore-unfixed: true
+
+      - name: Evaluate release image against CVSS policy (release gate)
+        if: ${{ inputs.build_type == 'production' }}
+        run: python3 .github/workflows/scripts/evaluate-trivy-report.py trivy-release-datacenter-${{ inputs.component_type }}.json
```
hatem-amairi-sonarsource, line 114: This scans sonarqube:<version>-datacenter-* after it is already public on Docker
Hub, and release.yml runs that promotion daily on a cron rather than only on
release. Since Trivy's database advances daily and the image can never change,
the first qualifying CVE in a released image turns this into a permanently red
nightly job. The only remediation available at that point is editing
.trivyignore.yaml on master, which would turn the suppression file into a rubber
stamp.

It also doesn't block much: the vulnerable image is already distributed, so
failing here mainly desynchronises Docker Hub from GCP Marketplace. Gitar covered
the expiry-date angle; the immutability and daily-cron angle is separate.

.github/workflows/push_and_pr.yml
```diff
@@ -199,13 +199,97 @@ jobs:
...
+
+      - name: Run Trivy vulnerability scanner (SARIF, for GitHub Security tab)
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        env:
+          TRIVY_SHOW_SUPPRESSED: 'true'
+        with:
+          image-ref: '${{ needs.load-config.outputs.staging_image }}:${{ needs.load-config.outputs.current_version }}-master-${{ matrix.edition }}'
+          format: 'sarif'
+          output: 'trivy-${{ matrix.edition }}.sarif'
+          severity: 'LOW,MEDIUM,HIGH,CRITICAL'
```
hatem-amairi-sonarsource: The JSON steps set TRIVY_SEVERITY including UNKNOWN; the SARIF steps pass
severity: 'LOW,MEDIUM,HIGH,CRITICAL'. Since 055bc4b an UNKNOWN finding blocks the
build but never reaches the Security tab, so the operator has no artifact to act
on. That works against the visibility goal that prompted the UNKNOWN change.

Adding UNKNOWN to the SARIF severity list would keep the two in step.

.github/workflows/push_and_pr.yml
```diff
@@ -232,23 +316,45 @@ jobs:
...
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        env:
+          TRIVY_SHOW_SUPPRESSED: 'true'
         with:
-          image-ref: '${{ needs.load-config.outputs.staging_image }}:${{ needs.load-config.outputs.current_version }}-master-datacenter-app'
+          image-ref: '${{ needs.load-config.outputs.staging_image }}:${{ needs.load-config.outputs.community_build_version }}-master-community'
           format: 'sarif'
-          output: 'trivy-datacenter-app.sarif'
+          output: 'trivy-community.sarif'
           severity: 'LOW,MEDIUM,HIGH,CRITICAL'
```
hatem-amairi-sonarsource: Same issue here

.github/workflows/scripts/tests/fixtures/unexpired_suppression.json
```diff
@@ -0,0 +1,24 @@
+{
+  "SchemaVersion": 2,
+  "ArtifactName": "example/image:test",
+  "Results": [
+    {
+      "Target": "example/image:test (debian 12.5)",
+      "Class": "os-pkgs",
+      "Type": "debian",
+      "Vulnerabilities": [],
+      "ModifiedFindings": [
```
hatem-amairi-sonarsource: Minor. Real Trivy populates ExperimentalModifiedFindings - 77 entries on both
0.70.0 and 0.74.0, with ModifiedFindings empty. The script handles both via
SUPPRESSED_FIELDS on line 61, so there's no runtime bug here, but every fixture
uses ModifiedFindings, which means the suppression-display path is only ever
exercised against a shape production doesn't produce.

Switching one fixture to the experimental key would cover the path that actually
runs in CI.

hatem-amairi-sonarsource (commented):
Hello @jimil09 ,
Good work, I added you few comments also I noticed no workflow invokes pytest, so the tests pinning the gate's scoring policy never
execute. A small job running `python3 -m pytest .github/workflows/scripts/tests`
would close it.

---
Not a blocker, and not a request to change this PR - but it may be worth checking (or adding a hardening ticket for it) before the script becomes something we maintain.

The ticket's premise is that the scanner's severity input filters on the
distribution's own priority rating rather than CVSS. That's true of the default
(auto), but Trivy does expose a native override:

    --vuln-severity-source nvd,redhat,redhat-oval,debian,ubuntu,alpine,...
    --severity MEDIUM,HIGH,CRITICAL,UNKNOWN
    --exit-code 1

Three flags, no script, gated by Trivy's own exit code - which is closer to what
ticket item 3 originally asked for. It emits a per-finding SeveritySource field,
so "which source was used" is reported natively, and it bands using the standard
CVSS ranges, which would make findings [2] and [3] above moot.

Two caveats found while testing:

  - --vuln-severity-source nvd on its own is unsafe. It doesn't relabel, it drops
    findings NVD has no data for: 185 down to 49 on an unfiltered scan of the
    test image, with "WARN No severity found in specified sources" for the rest.
    Fresh CVEs that only the distro has triaged would silently vanish instead of
    failing safe. The full fallback chain is required.

  - It's a different algorithm. Rule 2 in the ticket is "highest score across the
    remaining sources"; this flag is "first source in the list that has data".
    Provenance is reported at vendor granularity, not the CVSS version (v4/v3/v2)
    the ticket asks to record.

Rego (--ignore-policy) was also tested. The policy input does include the full
per-source CVSS map, so the ticket's exact precedence is expressible, and a test
policy reproduced it correctly - it removed the CVE-2025-69644/binutils family,
which Trivy labels MEDIUM but whose only CVSS data is redhat V3Score 2.8, exactly
the divergence this PR exists to catch. --exit-code 1 fired correctly and it
composes with --ignorefile. The problem is auditability: rego-filtered findings
largely vanish without a trace. Standalone, ExperimentalModifiedFindings was
absent entirely; combined with the ignore file, only 1 of roughly 27 rego-removed
findings got a "Filtered by Rego" entry.

Custom --compliance specs can't express a numeric CVSS threshold, --format
template is output-only and can't drive the exit code, and VEX / --ignore-status
filter on fix status rather than score.

So nothing native reproduces the ticket's spec exactly, and keeping the script is
defensible. But the spec was likely written without --vuln-severity-source in
view. The trade - coarser provenance and fixed-priority selection, against
roughly 450 lines less to maintain and two bug classes avoided - is worth an
explicit decision.

.trivyignore.yaml governance and expiry are unaffected either way; --ignorefile is
orthogonal and composes with all of the above.

jimil09:
@hatem-amairi-sonarsource all 8 findings from your review are fixed and pushed (`1997816`, `9988ea0`, `93530cd`). Thanks for the thorough review!

hatem-amairi-sonarsource (changes requested):
Hello @jimil09,
Sent you more issues privately. Feel free to look at them and assess if they are worth tackling in this PR :)

jimil09:
@hatem-amairi-sonarsource you're right, and I've reverted (`1b24086`). Full disclosure: I couldn't pull up either doc myself to verify firsthand, but the argument holds on what we've already confirmed in code — the promote step is mechanically just a DockerHub-to-GCR sync, so skipping it on schedule would have silently broken that. And treating a later-red nightly scan as something `.trivyignore.yaml`'s SLA governance handles (rather than something the gate should dodge) is consistent with why we built that governance model in the first place. Thanks for digging into the docs before flagging it.

jimil09:
Both fixed in `2c4ec7d`: each of the 6 glibc CVEs now has its own accurate justification instead of the shared disjunctive statement, and the list is re-sorted lexicographically.

One correction to the suggested fix text, for the record: CVE-2026-6791's upstream description says the unbounded allocation in `parse_tilde` uses `strndupa` (a stack-based allocation), not a heap allocation — the bug is a stack clash, not a heap overflow. Wrote it up as such. Verified all 6 statements against Trivy's actual upstream `Description` field (not just `Title`/CVSS vector) before writing them, and re-confirmed against a fresh Trivy DB scan that the suppressions still resolve CI.

hatem-amairi-sonarsource (approved):

hatem-amairi-sonarsource (approved):

